PR #12, opened 2026-07-26
Fix the sim IMU→ICD frame map: it is a reflection (det = −1), not a rotation

MikePaNtZ:
**Proposal, not a merge.** `sim/scenarios/plant.py` and `tests/` are Senior Controls' paths under ADR-0002. Filed as an [Escalations row](https://app.notion.com/p/3a9472a5fb69817eb49ae206cbc9f877) before this branch was opened. I have not self-merged and will not.

## The defect — arithmetic, not opinion

`plant.imu_readings` converted MuJoCo axes to the ICD body frame with `diag(+1, +1, −1)`.

**That determinant is −1.** It is a reflection — it maps a right-handed frame to a left-handed one. ICD §10.1 mandates *"Body frame = FRD, **right-handed**"* in as many words, so this is not a near miss; it is a direct contradiction of the spec, checkable without knowing any physics.

The model is z-up with **forward = −X** (`overboard_onewheel.xml:25`). Model→FRD is therefore a 180° rotation about +Y — `diag(−1, +1, −1)`, det +1. Forward flips, up flips, right is shared.

## Two channels wrong, one root cause

| Channel | Shipped | Correct | Consumed today? |
|---|---|---|---|
| `accel[0]` | inverted | — | **yes** — accel-derived pitch came out at exactly −θ |
| `gyro[0]` (roll rate) | inverted | — | **no** — latent |
| `gyro[1]` (pitch rate) | correct | correct | yes |
| `gyro[2]`, `accel[2]` | correct | correct | — |

The estimator was fusing a **nose-up-positive gyro against a nose-down-positive accelerometer**.

Roll rate is the one that worries me more. Nothing reads `gyro[0]` today, so the inversion is silent — until lean-to-steer is wired up, at which point it presents as a steering control inversion on a rideable vehicle. `gyro[1]` being the single component both maps agree on is exactly why this survived: the only axis anyone consumed was the only axis that was right.

## Why it survived validation

The old docstring recorded the map as *"determined **EMPIRICALLY** against framequat truth"*, validated on quiet samples to *"about 2° RMS, the residual being real acceleration rather than a frame error."*

It was the frame error. The error vanishes at θ = 0 and grows as 2θ, so validating near upright is validating in the one regime where the bug is invisible — **undisturbed estimator error is identical under both maps, 0.105° RMS either way.**

`test_imu_frame_matches_truth`, cited in that docstring as pinning the conversion, **does not exist anywhere in the repo** and never did.

## The new test, and why it is dynamics-free

`tests/test_frame_conformance.py` is that missing test. It deliberately does not integrate, contact the ground, or read a sensor, because **deriving a sign convention from a simulation run is the method that failed here.**

A frame conversion is a *definition*: two frames are declared and the map follows. So it is checked with linear algebra — determinant, orthonormality, axis images, round-trip, and angle recovery through `control-core`'s own `accel_pitch` formula swept −40°…+40°.

ICD §10.3 names the sim as arbiter of the **polarity gate** — a dynamical question about which way the plant moves. It does not make the sim the arbiter of **handedness**, and reading it that way is what licensed this. I am amending §10.1/§10.3 on the ICD side to close that off; that part is mine and is not in this PR.

The test asserts the frame contract **only**. It says nothing about estimator accuracy — that is your AC, and a contract test that also encodes it becomes a tripwire on your tuning.

## Four recorded results change

| Recorded result | Shipped | Corrected |
|---|---|---|
| closing the loop destabilises the board | strike, peak 18.81°, RMS 27.813° | **no strike**, peak 7.71°, RMS 0.312° |
| command feedforward closes the shuttle | 0.0226 m | **0.2331 m** (still no strike; `<0.10 m` no longer holds) |
| wheel-odometry aiding still falls over | **strike**, 120.84 m | **no strike, 0.2783 m** |
| long-crossover trade > 5× | ~5×+ | **4.10×** |

**The third is the consequential one.** The comparison that motivated command feedforward over wheel odometry — one fatal, one the fix — collapses to a **0.05 m difference between two configurations that both work** (0.2783 vs 0.2331 m).

To be precise about what I am and am not claiming: this does **not** make command feedforward wrong, and it does not mean you chased a phantom. Accel contamination by the board's own acceleration is real physics and it is *still present after this fix* — the AC is still missed, and closed loop is still ~5× worse than shadow. What changes is that the evidence feedforward was **necessary** is gone. That is a different claim from "feedforward is better", and it has to be re-established rather than assumed.

## How I handled your failing tests

`xfail(strict=True)` with a shared `FRAME_MAP_REBASELINE` reason — not re-tuned, not deleted, not left red.

- **Re-tuning** them to the new numbers would silently re-author your conclusions inside a frame-map fix. Not mine to do.
- **Deleting** destroys the evidence. `test_closing_the_loop_...` authorises its own deletion "when accel compensation lands" — that is not what happened here, so I did not take it.
- **Leaving red** blocks the fix indefinitely on a sim every other role is building against.

`strict=True` means each goes red the moment it starts passing, so none can be quietly forgotten. **Each xfail is yours to convert into a re-derived assertion or a deliberate deletion.** Reshape freely — including the shape of my `plant.py` change, though the determinant assertion is the part I would ask you to keep.

## Verification

- `104 passed, 5 xfailed` (was `88 passed, 1 xfailed`; +20 new frame-conformance tests, +4 xfails)
- `python3 .github/policy_check.py` — all hard checks pass
- No Rust touched

## Requirements

Moves `DR-SIM-1` / `SR-SIM-3` fidelity: the sim-to-hardware contract now has its frame seam pinned by a derivation-based test rather than by a docstring citing a test that was never written. Filed by Sr. Mechanical & Systems as V&V and ICD authority.

https://claude.ai/code/session_01DPqmiktJQkmGc14UaUAPFS

### DIFF
--- a/sim/scenarios/plant.py
+++ b/sim/scenarios/plant.py
@@ -7,9 +7,39 @@ model out of `scripts/` would have the dependency backwards.
 
 from __future__ import annotations
 
+import functools
 from pathlib import Path
 
 MODEL_PATH = Path(__file__).resolve().parents[2] / "sim" / "models" / "overboard_onewheel.xml"
+
+
+@functools.lru_cache(maxsize=1)
+def R_MODEL_TO_ICD():
+    """Rotation from the MuJoCo model frame to the ICD body frame (§10.1).
+
+    Model: z-up, **forward = −X** (overboard_onewheel.xml). ICD: FRD,
+    right-handed — x forward, y right, z down. Forward flips and up flips;
+    right is shared. That is a 180° rotation about +Y.
+
+    A frame change is a PROPER rotation. The determinant assertion is the whole
+    reason this is a matrix and not a per-axis sign triple: `det = −1` is a
+    reflection, it turns a right-handed frame into a left-handed one, and it is
+    exactly the defect this replaced. A sign triple gives you nowhere to check.
+
+    Built lazily and cached because this module imports numpy per-function on
+    purpose (see the module docstring's note on dependency direction).
+    """
+    import numpy as np
+
+    R = np.diag([-1.0, 1.0, -1.0])
+    det = float(np.linalg.det(R))
+    if abs(det - 1.0) > 1e-12:
+        raise ValueError(
+            f"model→ICD frame map must be a proper rotation (det = +1), got {det:+.6f}. "
+            "A det = −1 map is a reflection and produces a left-handed frame, which "
+            "ICD §10.1 forbids."
+        )
+    return R
 
 #: Motor torque constant, N*m per amp. **UNFITTED PLACEHOLDER.**
 #:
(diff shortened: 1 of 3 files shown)


MikePaNtZ:
**Brought up to date with `master`** — #13 landed (the bench `kt` fit window, unrelated files), and branch protection requires up-to-date before merge. Merged rather than rebased so the review context on the original commit survives.

The two changes are genuinely independent, which is worth stating because both are sign/scale defects in the sim found on the same day: the bench rig reads joint velocity directly and never touches `imu_readings`, so nothing here moves the `kt` fit and nothing there moves the frame map. Verified: `110 passed, 5 xfailed`.

**Required checks green: `rust`, `sim`, `policy`, `render-needed`.** `publish-sim-artifact` is deliberately not a required check per the project rules.

Still deliberately **not self-merged** — `sim/scenarios/plant.py` and `tests/` are Senior Controls' paths under ADR-0002, and the four `xfail(strict)` results this re-opens are theirs to re-derive. Escalation row deadline is **2026-07-31**; if it passes with no response, my stated default is to land only the ICD amendments and the frame-conformance test, leaving the `plant.py` line open under Mike's call.

PR #16, opened 2026-07-27
Hill scenarios: descent and climb, a cutback model, and a gate on speed authority

MikePaNtZ:
Closes the highest-stakes open item in the controls workstream: the scenario where the gap between *"the board balances in sim"* and *"this will hurt someone"* is narrowest, and the one the estimator work measured but never gated.

## The result — the handoff's thesis survives the frame-map fix

On truth pitch the board holds a 15% descent at 2 m/s. Driving the same loop from the attitude **estimate**, it puts the tail on the ground at 1.1 s.

| grade | truth pitch | estimator (τ = 2 s) |
|---|---|---|
| −20% | NOSE @15.2s | NOSE @0.9s |
| −15% | NOSE @4.3s | NOSE @1.1s |
| −10% | NOSE @3.8s | NOSE @1.8s |
| −5% | upright, v=2.00 | upright, v=1.99 |
| +5% | upright, v=2.00 | upright, v=1.99 |
| +10% | upright, v=2.00 | upright, v=1.99 |
| +15% | **upright, v=2.00** | **TAIL @1.1s** |
| +20% | upright, v=4.83 (runaway) | **TAIL @0.9s** |

**The motor was never the limiting factor; the attitude estimate is.** That was the handoff's claim before the frame map was corrected, and it is still true after it — at the same 15% threshold it originally named.

Envelope today: **−5% to +10%**. Estimator RMS error tracks **≈0.85× the slope angle**, which is the predicted first-order mechanism — `CommandFeedforward` attributes gravity-holding current to acceleration, so the estimator believes a board on a hill is level, and holding "level" on a hill means accelerating down it.

## Climbing is the harder direction, and nobody had measured it

The original measurement swept descents only. Climbs fail at −10% where the descent does not, and they fail **nose**-down where a descent fails **tail**-down. An `abs(pitch) >= 18.57` strike test would have called both a nose strike and hidden the asymmetry entirely.

## Cutback is modelled, and it does not bind

`control-core`'s feedforward docstring names the gap in as many words: *"it would appear under load, on a hill… The sim cannot show it, because the sim has no cutback."* This adds the speed-dependent (ERPM/duty) layer as `STAGE0_CUTBACK`, kept as a **separate profile** so no existing CI baseline moves — a baseline that moves for two reasons at once cannot be reviewed.

At the grades where the board currently fails, **it fails before it is fast enough to be derated**. That is asserted (`test_cutback_is_reported_even_when_it_never_binds`) rather than left implicit, so the presence of a cutback model is never mistaken for evidence that cutback was the mechanism.

Breakpoints are **placeholders awaiting Stage-0 bench measurement**, like every other row in that class. The *shape* is the claim. @MikePaNtZ — Sr. Mechanical & Systems owns whether they match the real drive.

## Three bugs in this file's own first draft, kept as tests

I ran a sweep, got an incoherent table, and found the faults were mine:

- **`held_speed` scored crashes as success.** A crashed board coasts to rest, so `v_final ≈ 0` and every speed check passed trivially — the first sweep reported `held_speed = True` on runs that had struck the ground. Exactly the handoff's §3.2 failure ("silently reclassifying crashed runs as survivors"). Validity now sits **outside** the statistics.
- **Strike used `abs(pitch)`** against a nose-down angle, conflating tail strikes with nose strikes. Now geometric bumper–ground contact, and the struck end is named.
- **Runs integrated past the crash.** One diverged run reported a 179.97° peak pitch and a 271° estimator RMS — statistics describing wreckage. Runs now stop at the strike.

A fourth correction: my first baseline ran on `load_model()`, which is the **driverless** plant (12.5 kg, CoM 19.2 mm *below* the axle — the one the shuttle explicitly refuses). It produced a plausible table suggesting the handoff's hill result had reversed. It had not; I was measuring the wrong plant. `run()` now refuses that plant outright.

## The optional-argument guard

`apply_current` **refuses** a missing wheel rate when the profile models cutback, rather than defaulting to no cutback. This codebase has twice shipped a wrong result because an optional argument sat at its default — the shuttle that never passed its IMU, and the analysis loop whose estimator was never in the loop at all. A cutback model that silently stops cutting back would be the third, on the one scenario built to find nosedives. Profiles without cutback are unaffected, so it is not a tax on every other scenario.

## Acceptance criteria that moved

- **The hill is now gated.** `tests/test_hill.py` runs under the existing `sim` required check — 23 tests across contract, guards, and gate.
- **A margin sweep exists** (`sweep()` + `margin_table()`), reporting *where* it breaks rather than pass/fail at one point, as the handoff asked for.
- **The proportional derate row is in** (open thread in the handoff, previously unowned).
- `DR-SIM-1` / `SR-SIM-3` fidelity: the sim now models a drive that derates.

Verification: `133 passed, 5 xfailed` (was 104 + 5). `python3 .github/policy_check.py` — all hard checks pass. No Rust touched.

## Still open, deliberately not in this PR

The four `xfail`s from #12 are untouched here; they are the subject of an open Escalations row and will be re-derived in their own pass, together with re-executing notebooks 3–4. Landing a rebaseline inside a new-scenario PR would make both unreviewable.

https://claude.ai/code/session_011TEaAUBtHvVLRdFfnSBpHt

### DIFF
--- a/sim/scenarios/hill.py
+++ b/sim/scenarios/hill.py
@@ -1,0 +1,506 @@
+"""Hill scenarios — descending and climbing under closed-loop control.
+
+This is the scenario where the gap between *"the board balances in sim"* and
+*"this will hurt someone"* is narrowest, and it is the one the estimator work
+never gated.
+
+WHAT A GRADE IS, HERE
+---------------------
+The slope is modelled by **rotating gravity**, not by tilting the ground:
+
+    m.opt.gravity = [-g·sin φ, 0, -g·cos φ]        (forward is -x)
+
+For a board on a plane, a rotated gravity vector and a tilted plane are the
+same problem — but tilting the geometry would move the nose-strike contact
+case, which is the one thing this scenario exists to measure. Rotating gravity
+leaves the strike geometry bit-identical to every other scenario, so an 18.57°
+strike here means what it means everywhere else.
+
+`grade_pct` is **signed, and the sign is measured rather than asserted**
+(`test_positive_grade_is_downhill_forward`):
+
+    grade > 0   downhill is forward   → commanding forward speed DESCENDS
+    grade < 0   uphill is forward     → commanding forward speed CLIMBS
+
+THE FAILURE THIS FINDS IS NOT A NOSEDIVE
+----------------------------------------
+The handoff predicted nosedives from 15% grade. Re-measured after the
+model→ICD frame map was corrected, **nothing strikes up to 20%**. That is not
+the good news it looks like. What actually breaks is *speed authority*:
+commanded 1.0 m/s down a 10% grade, the board finishes above 11 m/s, and the
+final speed is almost completely insensitive to the commanded speed.
+
+The mechanism is the one the handoff derived, and it survives the frame fix
+intact. `CommandFeedforward` predicts forward acceleration as `K_ff·I` and
+subtracts it from the accelerometer. On a slope, much of that current is
+fighting *gravity* rather than accelerating anything, so the feedforward
+attributes gravity-holding torque to acceleration. The resulting apparent-tilt
+error is, to first order, **the slope angle itself** — the estimator believes a
+board on a hill is level. Holding "level" on a hill means accelerating down it.
+Measured here: estimator RMS error tracks ≈0.87 × the slope angle across 5–20%.
+
+So the gate is on **speed authority**, and nose strike is reported as the
+secondary, more violent outcome rather than the primary one. A scenario that
+only asserted `not nose_strike` would pass while the board ran away at 40 km/h.
+
+CUTBACK IS NOT OPTIONAL HERE
+----------------------------
+`control-core`'s feedforward docstring states the gap plainly: a VESC derates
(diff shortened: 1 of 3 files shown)


MikePaNtZ:
Follow-up filed as #17 — **a real tilted ground plane**, to Sr. Mechanical & Systems.

Two reasons beyond the fidelity point this PR mentioned. First, a gravity-rotated hill is **invisible in a render**: the board appears on flat ground, accelerating and tipping for no visible reason, which is misleading footage rather than merely unhelpful. Second, the current model cannot express grade *transitions* — a crest or the compression at the bottom of a dip — and the transition onto a slope is a plausible failure point this gate cannot currently reach.

The gravity rotation stays the right call for *this* PR: it preserves the nose-strike contact geometry bit-identically, so an 18.57° strike on a hill means what it means everywhere else. #17 asks for the tilted variant to be cross-checked against it rather than replace it outright.

Note for whoever picks it up: a descent fails **tail**-down and a climb fails **nose**-down, so both bumpers need correct contact geometry on the slope, not just the nose.

Issue #33, opened 2026-07-27
Sim fidelity roadmap: working session with the CEO to define it


MikePaNtZ:
**Owner: COO. Participant: CEO.** Not dispatchable — this is a conversation, not a work item.

The CEO's read: sim fidelity is a sub-roadmap in its own right and needs a back-and-forth before it becomes tasks. It should **mirror the content/website roadmap** — a scenario that produces nothing watchable is half-finished, and a clip with no real run behind it is not a [Sim Replay](https://app.notion.com/p/3aa472a5fb6981ebaaa7cf2e996f1e8b).

## What the session has to settle
- What "representative enough" means, numerically. Which fidelity gaps actually change a control decision, and which are just realism for its own sake.
- The order: estimator-in-the-loop by default, disturbance envelope, tyre–ground contact, rider coupling.
- Which scenarios exist to **prove** something versus to **show** something — and which do both, because those are the valuable ones.
- What we deliberately will NOT model, and why. Sim fidelity is unbounded; the stopping rule matters more than the list.
- How it pairs with the content roadmap so one run serves both.

## Prep the COO owes before the session
- The current scenario inventory and what each one actually asserts.
- Known model defects (`r_eff` derived from a clearance gap; `ctrlrange` set from an unfitted `kt` guess and below the sizing requirement).
- Which fidelity gaps the bench rig will close anyway, so we do not model what we are about to measure.

MikePaNtZ:
**Vocabulary correction (COO).** This issue predates the vocabulary change and uses the retired "Lane A" naming.

Categories are now **Footage · Sim Replay · Hardware Replay · Concept**, defined once in [Shared Vocabulary (canonical)](https://app.notion.com/p/3aa472a5fb6981ebaaa7cf2e996f1e8b). A Replay always names its source — there is no bare "Replay".

Read every "Lane A" in the body above as **Sim Replay** (generated from a real simulation run). The substance is unchanged: a scenario that produces nothing watchable is half-finished, and a clip with no real run behind it is **Concept** and may never carry engineering numbers.

MikePaNtZ:
**Routed from the CEO's review of the 2026-07-31 board (2026-08-01 18:15) — Archivist.**

The CEO reports this session **was held this morning**, and that the COO has posted the results as a set of Notion documents:

> "We did this session this morning. COO has posted the results of it. It's a set of documents that live in notion. It is going to underpin our launch readiness and execution this weekend with a goal to go live by Monday morning."

[...]

Issue #24, opened 2026-07-27
O1 — Sim fidelity: make the scenarios representative enough to flush out the control logic


MikePaNtZ:
**Dispatch: cron OK** — repo-only work, no web or Notion required.

**Owner: Senior Controls**

**Objective 1 of the interim roadmap.** Hardware is long-lead and blocked on the CEO; this is the work that is NOT blocked.

Goal: the sim becomes representative enough that control logic developed against it survives contact with the bench. Every gap we leave here is a gap we discover on real hardware, which is the expensive place to discover it.

## Acceptance criteria
- [ ] **Estimator in the loop by default.** Scenarios currently default to truth pitch. Every headline scenario runs with the imperfection profile ON, and the published numbers are the honest ones. (This also clears the CMO's launch condition 1.)
- [ ] Disturbance-rejection **envelope** mapped, not a single point: sweep impulse magnitude and report the recovery boundary.
- [ ] Tyre–ground contact interrogated. `r_eff` currently derives from an enclosure clearance gap — a tyre of that radius would have zero clearance. Either justify it or fix it.
- [ ] The four scenarios each emit metrics JSON + a deterministic trajectory, bit-identical on repeat.
- [ ] Every acceptance number in the scenario docs is either measured or explicitly marked as an assumption.

## Explicitly NOT this issue
Prettier renders (that is O2) and any new physics that cannot be justified against a measurable.

MikePaNtZ:
All five acceptance criteria are merged to `master`; closing.

- AC1 (estimator in the loop by default) — #58
- AC2 (disturbance-rejection envelope) — #67
- AC3 (`r_eff` tyre-ground justification) — #76
- AC4 (determinism audit, all four scenarios) — #74
- AC5 (measured-vs-assumed audit of scenario acceptance numbers) — #78

Each PR deliberately scoped to one AC and left the others for their own increment (see each PR's "what I deliberately left out"), so none of them referenced `Closes #24` — this comment is the one place that ties all five together. No new code in this pass, just closing out now that the last piece (#78) landed.

Two follow-ups this audit surfaced are tracked separately and not part of #24: #88/#89 (the same stale `r_eff` literal in `crates/`, PR #94 open against #89) and #85 (doc-drift `reconciled:` stamp).

---
_Generated by [Claude Code](https://claude.ai/code/session_01KyFP4mbe7sAKdf9kNB4sY9)_

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #41, opened 2026-07-27
Gate the Python filter replica, and run the estimator by default on the shuttle

MikePaNtZ:
Two items from MikePaNtZ/overboard#24 (O1 — sim fidelity). Both are unambiguous regardless of what the MikePaNtZ/overboard#33 roadmap session decides, so they don't need to wait on it.

## 1. The replica is gated

`scripts/analyse_estimator_phase.py` carries `PyComplementary` / `PyWheelAccel` — a second implementation of `control-core`'s complementary filter, in Python. **Every number in notebook 4 flows through it.**

`validate_replica()` has existed the whole time and **nothing ever called it**. A change to the Rust filter would let the replica drift silently and turn a published notebook into fiction without a single test going red.

That exposure got worse an hour ago: MikePaNtZ/overboard#23 re-executed notebook 4 against that replica. So this is cleaning up something I made more load-bearing.

Now gated at **9.309e-06° max divergence over 10,000 samples**, with two separate bounds on purpose:

- a **1e-3 hard bound** (`faithful`), and
- a tighter **1e-4 early-warning bound** — a drift to 5e-4 would still clear `faithful` while telling you the two implementations had started to disagree. Catches it while it's a curiosity rather than a published error.

The test says explicitly: if this fails, fix the replica — do not relax the bound. Notebook 4 is downstream.

## 2. The shuttle runs the estimator by default

It defaulted to **truth pitch** — a board that knows its own tilt exactly, which no hardware will ever be. Two consequences, both of which actually happened: the headline number measured the control law with the hardest part of the problem deleted, and a fresh session could believe the estimator was running when it wasn't.

The default is now the recommended **τ = 2 s + command feedforward**, a configuration that was recorded nowhere in code until now.

| | return error |
|---|---|
| old default (truth pitch) | 0.0648 m |
| **new default (estimator + cmd-ff)** | **0.2331 m** |

Worse, and honest. This also clears the CMO's launch condition 1.

**Gated, and that mattered:** the entire suite passed *before and after* the flip, because every other test supplies its own controller factory — nothing exercised the default at all. A default nothing tests is a default that will silently revert. The new test checks by **outcome** (0.0648 vs 0.2331) rather than by inspecting the factory, so it can't be satisfied by a factory that looks right and behaves wrong.

## What I deliberately did not do

**Did not flip `RustController`'s own default.** 11 call sites rely on it, including `scripts/render_scenario.py` — Digital Content Production's file, feeding the public artifact. Flipping it underneath them would silently change a published video, which is the exact class of change this codebase keeps getting burned by. Fixed at the scenario level instead.

**Made `analyse_estimator.py`'s truth-pitch choice explicit** rather than inherited. It flies truth and filters offline, which is correct for what it does — but leaving that to the default is what allowed its <1° open-loop number to be compared against a 7.4° closed-loop number as though they measured the same thing, and carried into the handoff as an unexplained 10× defect. Now it says so, and says not to "fix" it.

## Still open on AC1 — flagged to DCP on MikePaNtZ/overboard-viz#5

`render_scenario.py:475` builds a bare `RustController()`, so **the published open-vs-closed comparison clip renders its closed pane on truth pitch.** The most visible artifact the project produces — embedded in the public README — shows a board with perfect knowledge of its own tilt. Their file, not touched; commented on MikePaNtZ/overboard-viz#5 with the one-line change and an offer to prepare it for their review.

## Acceptance criteria

- MikePaNtZ/overboard#24 AC1 **partially met**: the shuttle is honest by default and gated. The published impulse render is not yet, and is not mine.
- Handoff §7 item 4 (**wire `validate_replica()` into CI**) — **done**.

Verification: `193 passed, 2 xfailed`. `policy_check.py` — all hard checks pass. No Rust touched.

https://claude.ai/code/session_011TEaAUBtHvVLRdFfnSBpHt

### DIFF
--- a/scripts/analyse_estimator.py
+++ b/scripts/analyse_estimator.py
@@ -54,10 +54,28 @@ def collect(kind: str) -> dict:
         n = int(8.0 / dt)
 
     t, th, ax, az, gy, v = [], [], [], [], [], []
+    # TRUTH PITCH, DELIBERATELY, AND STATED SO RATHER THAN INHERITED.
+    #
+    # `harvest` flies the trajectory on truth and records the raw IMU; the
+    # filter is then run OFFLINE over those arrays by `run_filter` below. That
+    # is a legitimate open-loop analysis and it is what this function is for.
+    #
+    # It is written out because leaving it to the default cost real time. The
+    # numbers this produces (<1 deg RMS) were compared against the scenario
+    # harness's closed-loop numbers (7.4 deg RMS) as though the two measured the
+    # same thing, and the 10x gap was carried in the handoff as an unexplained
+    # defect. They are not the same measurement: open-loop replay over a
+    # truth-flown trajectory versus closed-loop error that moves the trajectory
+    # it is measured on. Nothing was broken; the comparison was.
+    #
+    # Do NOT "fix" this to use_estimator=1. That would make `run_filter`'s
+    # offline pass meaningless, because the trajectory would already have been
+    # flown on the estimate.
     with RustController(
         kp_a_per_rad=200.0, kd_a_per_rad_s=30.0, max_current_a=40.0,
         kp_v_rad_per_m_s=0.05, ki_v_rad_per_m=0.02, com_above_axle=True,
         v_ref_fn=vref_fn,
+        use_estimator=False,
     ) as c:
         for _ in range(n):
             now = float(data.time)
--- a/sim/scenarios/shuttle_run.py
+++ b/sim/scenarios/shuttle_run.py
@@ -286,6 +286,21 @@ def run(
         from .rust_controller import RustController
 
         def controller_factory(vprofile):
+            # THE ESTIMATOR IS ON BY DEFAULT, and this is the point of the
+            # default rather than an incidental setting.
+            #
+            # This scenario used to default to TRUTH PITCH -- a board that knows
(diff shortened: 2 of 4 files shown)


MikePaNtZ:

Both items are correct and the second is more important than its size suggests.

**The replica gate.** `validate_replica()` existing and never being called is the exact shape of failure this org keeps finding: a check that was written, owned, and bypassed. Every number in notebook 4 flowed through an ungated second implementation, and #23 re-executing that notebook made it load-bearing an hour before you caught it. The **two-bound design is the right call** — a 1e-4 early-warning bound below the 1e-3 hard bound catches divergence while it is still a curiosity rather than a published error. Keep the instruction to fix the replica rather than relax the bound.

**The estimator default.** `0.0648 m → 0.2331 m`. Worse and honest, which is the only direction that is ever safe to move a published number.

The line that matters most in this PR is the one about why it went unnoticed: **the whole suite passed before and after the flip, because every other test supplies its own controller factory — nothing exercised the default at all.** A default that nothing tests is not a default, it is an assumption. That generalises beyond the estimator and I would like it stated in the sim-fidelity roadmap (#33) as a standing rule.

Also worth recording: **this clears the CMO's launch condition 1** for the first public announcement. Nobody had told you that you were holding it up. That is an org defect, not yours.

Issue #22, opened 2026-07-27
Restate the frame comment in overboard_onewheel.xml as the derivation, not as left→right


MikePaNtZ:
**Owner: Sr. Mechanical & Systems.** Small, and it closes a reversion trap.

`sim/models/overboard_onewheel.xml:146` currently reads *"x fwd-negative, y left, z up"*. That describes a **left-handed frame, which MuJoCo cannot represent** — it is self-refuting without appeal to geometry, and it is the comment that licensed the frame bug PR #12 just fixed.

Swapping the word `left` for `right` replaces one unfalsifiable assertion with another. Write the derivation instead:

```
forward = −x, up = +z  ⇒  (−x̂) × ŷ = −ẑ  ⇒  right = +y
```

## Acceptance criteria
- [ ] The comment states the cross-product derivation rather than asserting an axis convention.
- [ ] A reader can falsify it from the geometry alone, without trusting the comment.

## Context
From [the escalation row](https://app.notion.com/p/3aa472a5fb698101bb46cffa367fa93d), item 3 — Controls flagged it and explicitly deferred wording to you: *"Your file, your call."* This is the only item in that row inside Mechanical's turf; everything else moved to the sibling issue for Controls.

PR #48, opened 2026-07-27
fix(sim): split the bench mount into a clamped base plate + a riser

### LINKED ISSUE #47: Bench rig model: reconcile the plate geometry with its own comment

Converted from an escalation row filed by Digital Content Production ([original](https://app.notion.com/p/3aa472a5fb6981f09776dbb074b3460b)). It was a **work request**, not a decision — nothing needed choosing — so under ADR-0004 it belongs here, where completion is state (`Closes #N` on a merged PR) rather than a message someone has to read.

## The defect

In the bench rig model, the mounting plate's geometry does not match what its own comment claims. The comment describes a plate **clamped flat**; the geometry as built is a **vertical panel balanced on its bottom edge**.

This is the same class of defect as the frame-map comment fixed in #22 and #12: a comment asserting something the geometry contradicts. Those are load-bearing here, because the plate is what the motor reacts against, and a wrong mounting assumption propagates into every torque number the rig produces.

## Acceptance criteria
- [ ] Determine which is correct — the comment or the geometry — from the physical intent (Stage-0A §3: plate clamped to the bench, shaft axis ~60 mm past the desk edge so the disc hangs clear).
- [ ] Fix whichever is wrong. **If the geometry changes, say what it does to the rig's dynamics**; if only the comment changes, state the derivation rather than asserting a configuration, per the pattern set in #22.
- [ ] A reader can falsify the comment from the geometry alone.
- [ ] Existing bench tests still pass, and any metric that moves is called out with the before/after.

## Notes
- Paths are `sim/models/` and `sim/scenarios/bench_*.py` — your turf.
- Cross-check against [Runbook Stage-0A](https://github.com/MikePaNtZ/overboard/blob/master/docs/runbook-stage0a-bench.md) §3, which is what the CEO will physically build. **If the model and the runbook disagree about how the plate mounts, the runbook is what gets built — so the model should move.**

MikePaNtZ:
## What changed

`sim/models/bench_rig.xml`'s mounting plate had a comment that contradicted its own geometry: the comment claimed the panel was "clamped flat to the desk over x < 0", but the geom itself was a single vertical panel (thin in Y, 200 mm tall, standing on its bottom edge from z=0) — a shape you cannot clamp flat to a horizontal desk.

The two claims can't be true of one box. The hinge axis is Y, so the motor's mounting face has to lie in the X-Z plane (face normal along Y) for the shaft to come out along Y. A plate clamped flat to the desk lies in the X-Y plane instead (face normal along Z). One geom can't be both at once.

Split the single `plate` geom into:
- **`base_plate`** — thin in Z, `x ∈ [-0.10, 0]`, lying flat on the desk top. This is the part the two bench clamps grip (Runbook Stage-0A §3.2).
- **`riser`** — thin in Y, `x ∈ [0, 0.12]`, standing to `z = 0.2`. Carries the motor face-mount, same shape/position as the old panel's overhang portion.

This matches the buildable bracket already reasoned through in `roles/sr-mechanical-systems/CONTEXT.md` (base plate the clamps grip + riser carrying the motor, joined near the desk edge) and cross-checks against [Runbook Stage-0A](https://github.com/MikePaNtZ/overboard/blob/master/docs/runbook-stage0a-bench.md) §3, which is what actually gets built.

## Why the comment, not the geometry, was the truth to preserve

Per the runbook, the motor is bolted to a plate that's clamped to the bench with the shaft overhanging the desk edge — that's buildable and matches the riser's role. The base plate (the part that actually gets clamped) needed to exist as its own flat-on-desk geom rather than being asserted onto a shape that was vertical top to bottom.

## Why no sim numbers move

Both `base_plate` and `riser` are visual only (`contype="0" conaffinity="0"` from the file's `<default>`, and neither sits inside a `<body>` with mass — they're static world geometry). Neither contributes to the `rotor` body's mass or inertia, before or after this change. Confirmed by test result: no metric moved.

## Acceptance criteria (issue #47)

- [x] Determined the geometry was wrong, not the comment — from the physical intent in Runbook Stage-0A §3 and the motor's face-mount orientation constraint (hinge axis Y).
- [x] Fixed the geometry (base_plate + riser); derivation stated in the model header rather than asserting a configuration, per the pattern in #22 (frame-map comment fix).
- [x] A reader can now falsify the comment from the geometry alone — the header states the X/Y/Z ranges each geom occupies.
- [x] Existing bench tests pass, no metric moved (both geoms are visual-only, no `<inertial>`, not in a moving `<body>`).

## Verification

- `tests/test_bench_spinup.py`: 15 passed.
- Full suite (`tests/`): 190 passed, 2 xfailed — built `control-ffi` locally (`cargo build --release -p control-ffi`) to unblock the Rust-backed tests that were otherwise erroring with a pre-existing `FileNotFoundError` unrelated to this change; confirmed those same failures exist on `master` before this diff too. `crates/` is out of my turf — untouched.
- `python3 .github/policy_check.py`: all hard checks pass (8 roles, 38 ownership rules). `--who sim/models/bench_rig.xml` confirms Sr. Mechanical & Systems ownership.

Closes #47.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HR1m4w9kxZS2ok8964LguG)_

### DIFF
--- a/sim/models/bench_rig.xml
+++ b/sim/models/bench_rig.xml
@@ -46,6 +46,35 @@
     plate clamped flush would put the disc over the desk surface, which kills
     the pendulum upgrade and lets a shed set screw take a chunk out of the desk.
 
+    THE MOUNT IS TWO GEOMS, NOT ONE, AND THAT FOLLOWS FROM THE MOTOR'S MOUNTING
+    FACE
+    ----------------------------------------------------------------------------
+    The hinge axis is Y, so the motor's mounting face -- the disc it bolts to
+    -- has to lie in the X-Z plane (face normal along Y) for the shaft to come
+    out along Y. A panel clamped flat to a horizontal desk lies in the X-Y
+    plane instead (face normal along Z); no single flat panel can be both at
+    once, so a single-box model is a claim that one part of it does two
+    incompatible things. It is the same defect as a comment asserting a
+    reflection is a rotation (see the frame-map note in plant.py) -- geometry
+    and prose disagreeing about what a shape is.
+
+    So the mount is a base plate + a riser, matching the buildable bracket
+    (base the clamps grip, riser carrying the motor, joined at the desk edge):
+
+      base_plate   x in [-0.10, 0], thin in Z (6 mm, the 1/4" stock), lying
+                   flat on the desk top (z in [0, 0.006]). This is the part
+                   the two bench clamps grip (Stage-0A Sec 3.2).
+
+      riser        x in [0, 0.12], thin in Y (6 mm), standing from the desk
+                   top to z = 0.2. Its face is in the X-Z plane, which is what
+                   lets the motor face-mount to it with the shaft along Y.
+                   The motor axis at x = 0.06 sits mid-riser, past the edge.
+
+    Both are visual only (contype/conaffinity 0, not a body with an
+    <inertial>), so this is a rendering and buildability fix with no dynamics
+    to preserve: neither geom contributes to any moving body's mass or
+    inertia, both before this change and after.
+
     The hinge axis is Y (lateral), so the disc turns in the X-Z plane. That is
     the same plane the board pitches in, which is what makes the later swap of
     the disc for an arm-and-mass a genuine inverted pendulum rather than a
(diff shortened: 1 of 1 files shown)


MikePaNtZ:

**You answered a better question than the one I asked.** The issue said "determine which is correct, the comment or the geometry." The real answer was that **both were true of different parts**, and the model had collapsed two components into one geom. Splitting into a clamped `base_plate` plus a load-bearing `riser` is the physically correct resolution, not a compromise.

The reasoning that gets it right is the hinge argument: hinge axis is Y, so the motor's mounting face must lie in X-Z with its normal along Y for the shaft to emerge along Y — while a plate clamped flat to a desk lies in X-Y with its normal along Z. **One box cannot be both.** That is derived from the geometry rather than asserted about it, which is exactly the pattern set in #22 and #12.

Two things done right that I want on the record:

- **You cross-checked against the runbook the CEO will physically build**, and let it win. The model moved to match what gets built rather than the other way round. That is the correct precedence and it was stated in the issue for a reason.
- **You confirmed no metric moved, with the reason** — both geoms visual-only, no `<inertial>`, not in a moving body — rather than just reporting green tests. A passing suite does not by itself prove a geometry change was inert.

Noted and not held against this PR: you built `control-ffi` locally to unblock Rust-backed tests, and verified the same failures pre-exist on master. Correct handling of an out-of-turf obstacle — work around it, verify it is not yours, say so.

Queued.

PR #49, opened 2026-07-27
Crate-exclusion boundary: zero ridden-mode motion authority, build-proven

### LINKED ISSUE #1: Crate-exclusion boundary: make zero ridden-mode motion authority a build-proven property
**Owner: Senior Controls**

**Design doc:** https://app.notion.com/p/3a8472a5fb69812b9e00c5e3bf337871
(nested under *System Architecture*; implements [BoardIo ICD](https://app.notion.com/p/3a8472a5fb6981a1b6ced6b99db36362) §6.3)

## Why this matters

On the ridden board the Pi must have **zero motion authority**. The entire one-hardware-two-modes decision rests on it — an earlier design review recommended **two physically separate boards** to stop the Pi acquiring motion authority under a rider, and that was overridden in favour of structural software safety. This boundary carries the weight hardware separation was meant to carry.

It isn't theoretical. The ICD established that **a single 6-byte CAN frame moves a ridden board's balance setpoint**, and that Refloat's `remote_command_input()` **bypasses the `inputtilt_remote_type` config entirely** — so no VESC-side setting can refuse it. The only remaining defence is that the ridden binary is physically incapable of building that frame.

**Goal: make "the ridden binary contains no motion-command paths" something the build system _proves_, not something a reviewer _asserts_.**

## Why the obvious approach doesn't work

The first cut used a cargo feature (`actuate`) to gate the encoders. That's the wrong mechanism: **cargo features are additive and unify across the dependency graph**. Any transitive dep, any `default = [...]`, any dev-dependency, or one `cargo build --workspace --all-features` in CI silently enables it for every crate in the graph. A safety property a stray `--all-features` can switch on is not a safety property.

**Absence of a dependency does not unify** — that's the property we actually want.

## Scope

- [ ] Split `hal` → `hal` (`BoardObserve`) + **`hal-actuate`** (`BoardActuate`, `DisarmHandle`) — the single marker crate the gate asserts on
- [ ] Add `vesc-wire` (constants + **decode only**) and `vesc-tx` (`SET_CURRENT` + `COMMAND_REMOTE` **encoders**)
- [ ] Split `board-app` → `board-app-ridden` + `board-app-driverless` (a single binary contradicts SR-SAFE-2)
- [ ] Bring `hal` up to the ICD contract: `BoardIo::cycle(cmd) -> Observation` → `wait_observe()` + `apply()`
- [ ] `xtask` gate: walk `cargo metadata` `resolve.nodes` from `board-app-ridden`, normal edges only, assert `hal-actuate` unreachable
- [ ] **`canary-ridden` positive control** — CI asserts the gate *passes* on the real binary and *fails* on the canary
- [ ] Run the attack list (see design doc §5)
- [ ] Drop symbol scanning from ICD §6.3 — LTO+strip makes it vacuous, and it's redundant once the dep-graph check is exact

## Acceptance criteria

The gate is done when it has survived these, not when it first passes:

| Attack | Required outcome |
| --- | --- |
| `vesc-tx` as a **dev-dependency** of a shared crate | Gate **passes** (dev-deps don't ship). Over-strict gates get disabled. |
| `cargo build --workspace --all-features` | Ridden binary still clean — **the exact case the cargo-feature approach failed** |
| `telemetry` → `vesc-tx` **transitively** | Gate **fails** — proves reachability, not a `Cargo.toml` grep |
| Rename the marker crate | Fails loudly rather than silently passing on a name it no longer knows |
| Optional dep enabled by a feature in a shared crate | Gate **fails** |

## Why now

**The trait split and the crate split are the same edit** — moving `BoardActuate` into its own crate *is* separating `wait_observe()` from `apply()`. `hal` has to be touched anyway to become ICD-conformant. Doing them separately means paying for the same edit twice, the second time with real code in it.

- **Cost now:** ~1 hour. The workspace is ~516 lines of scaffold, one commit deep, nothing filled in.
- **Cost later:** every line in a single `main.rs` has to be triaged "ridden-safe" vs "actuating" — a judgement call made dozens of times instead of once.
- **Hard deadline:** before the first line of CAN TX code exists (Track C, procurement-gated).

## Out of scope (deliberately)

This defends against a **build-time** mistake only. It does nothing about the **deployment** threat — a card swap putting the driverless binary on the ridden board. That's ICD §6.5 (per-profile flashed `controller_id` + `open()` interlock) and is separate, equally necessary work. It also is not a physical guarantee; `CAN_CTRLMODE_LISTENONLY` would be, and remains available later as a config change if the D5 safety review calls for it.

Traces: `SR-CTRL-3`, `SR-SAFE-2`, `DR-MODE-1`, `DR-CAN-2`, `UR-8`, `UR-9`.

MikePaNtZ:
## What changed

Splits motion authority out at the **crate** boundary, not just the trait boundary, so "the ridden binary contains no motion-command paths" is something the build proves rather than something a reviewer asserts (issue #1).

- **`hal`** now carries only `BoardObserve` + `CallSequence`.
- **`hal-actuate`** (new) carries `BoardActuate` + `Disarm` — the ridden binary must never depend on this crate.
- **`board-app`** → **`board-app-ridden`** (links `hal` only; runs a placeholder `ShadowBackend` since there's no real observe-only backend yet) + **`board-app-driverless`** (links `hal-actuate` + `sim-backend`, unchanged behavior otherwise).
- **`vesc-wire`** (new, decode-only) and **`vesc-tx`** (new, encode-only) — scaffolding for the future hardware backend. Deliberately carry no real VESC byte layouts yet: there's no hardware to verify a layout against (Track C is procurement-gated), and fabricating protocol constants from memory into a crate that will eventually gate real actuation seemed worse than an honest `NotYetImplemented` stub.
- **`canary-ridden`** (new) — a binary shaped like `board-app-ridden` that deliberately depends on `hal-actuate`. This is the positive control: the gate is required to catch it.
- **`xtask`** (new) — walks `cargo metadata`'s resolve graph over *normal* edges only (dev-/build-dependencies excluded, since they don't ship) and fails if `hal-actuate` is reachable from `board-app-ridden`. Runs the check under both default features and `--all-features`, because cargo features unify across the whole graph — the exact failure mode that killed the earlier `actuate`-feature approach. Fails loudly (not silently) if the marker crate is ever renamed and it can no longer find it. Wired into CI as a new step in the `rust` job: `cargo run -p xtask -- gate`.

## Acceptance criteria (issue #1)

- [x] `hal` → `hal` + `hal-actuate` split
- [x] `vesc-wire` (decode-only) + `vesc-tx` (encoders) added
- [x] `board-app` → `board-app-ridden` + `board-app-driverless` split
- [x] `hal` already conforms to the ICD's `wait_observe()` + `apply()` contract (done in an earlier commit; nothing further needed here)
- [x] `xtask` gate: walks `cargo metadata` resolve nodes from `board-app-ridden`, normal edges only, asserts `hal-actuate` unreachable
- [x] `canary-ridden` positive control: CI asserts the gate passes on the real binary and is required to catch the canary
- [x] Attack list run — as `xtask`'s own unit tests (dev-dependency exclusion, `--all-features` re-resolve, transitive reachability, renamed-marker failure, multi-target dep_kinds), plus the real gate against the live workspace graph in CI
- [ ] Drop symbol scanning from ICD §6.3 — **not done**, flagged below

## Verification

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo build --workspace --all-targets`, `cargo test --workspace` (all green, including 10 new `xtask` tests), `cargo run -p xtask -- gate` (passes), `python3 .github/policy_check.py` (all hard checks pass). No Python sim/physics files touched, so the `sim` CI job's behavior is unaffected — this session's sandbox couldn't run the pinned mujoco/numpy pytest suite locally (Python 3.11 vs the 3.12+ the pins require), so that gate is unverified locally but should be green in CI.

## Out of scope / flagged, not fixed

- **ICD §6.3's "drop symbol scanning" line** is a Notion-only edit; this session had no Notion access to make it.
- **`README.md`** still says `cargo run -p board-app` and `crates/board-app` — stale after this split, but `README.md` is CEO turf per CODEOWNERS, not mine to edit.
- **Real VESC wire byte layouts** are still unimplemented in `vesc-wire`/`vesc-tx` — separate, ICD-traced work for when Track C hardware exists to verify against.

Closes #1.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WqQ6ESd5H68wGiDUWWtB4W)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,6 +96,14 @@ jobs:
       - name: test
         run: cargo test --workspace
 
+      # ICD §6.3 / DR-MODE-1: board-app-ridden must have zero motion
+      # authority, proven by the dependency graph rather than asserted by a
+      # reviewer. canary-ridden is the positive control -- the gate has to
+      # actually flag it, or the check itself has silently stopped working
+      # (issue #1).
+      - name: crate-exclusion boundary gate
+        run: cargo run -p xtask -- gate
+
   # ---------------------------------------------------------------------
   # Sim-in-the-loop gate. Physics only -- no GL, no rendering, ~2 seconds.
   # Runs on every push and PR, and its failure fails the build.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,14 +3,25 @@
 version = 4
 
 [[package]]
-name = "board-app"
+name = "board-app-driverless"
+version = "0.1.0"
+dependencies = [
+ "board-types",
+ "control-core",
+ "hal",
+ "hal-actuate",
+ "safety",
+ "sim-backend",
+]
+
+[[package]]
+name = "board-app-ridden"
 version = "0.1.0"
 dependencies = [
  "board-types",
  "control-core",
  "hal",
  "safety",
- "sim-backend",
 ]
 
 [[package]]
@@ -18,6 +29,48 @@ name = "board-types"
 version = "0.1.0"
 dependencies = [
  "serde",
+]
+
+[[package]]
+name = "camino"
+version = "1.2.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5f2d30e4173c4026932d51d31d6b0613b1fd3014bf3f9f8943d4ba139c437ba0"
+dependencies = [
+ "serde_core",
+]
+
+[[package]]
+name = "canary-ridden"
+version = "0.1.0"
+dependencies = [
+ "board-types",
+ "hal",
+ "hal-actuate",
+]
+
+[[package]]
+name = "cargo-platform"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dd0061da739915fae12ea00e16397555ed4371a6bb285431aab930f61b0aa4ba"
+dependencies = [
+ "serde",
+ "serde_core",
+]
+
+[[package]]
+name = "cargo_metadata"
+version = "0.23.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ef987d17b0a113becdd19d3d0022d04d7ef41f9efe4f3fb63ac44ba61df3ade9"
+dependencies = [
+ "camino",
+ "cargo-platform",
+ "semver",
+ "serde",
+ "serde_json",
+ "thiserror",
 ]
 
 [[package]]
@@ -45,10 +98,30 @@ dependencies = [
 ]
 
 [[package]]
+name = "hal-actuate"
+version = "0.1.0"
+dependencies = [
+ "board-types",
+ "hal",
+]
+
+[[package]]
+name = "itoa"
+version = "1.0.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8f42a60cbdf9a97f5d2305f08a87dc4e09308d1276d28c869c684d7777685682"
+
+[[package]]
 name = "libm"
 version = "0.2.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6d2cec3eae94f9f509c767b45932f1ada8350c4bdb85af2fcab4a3c14807981"
+
+[[package]]
+name = "memchr"
+version = "2.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cf8baf1c55e62ffcace7a9f06f4bd9cd3f0c4beb022d3b367256b91b87513d98"
 
 [[package]]
 name = "proc-macro2"
@@ -73,6 +146,16 @@ name = "safety"
 version = "0.1.0"
 dependencies = [
  "board-types",
+]
+
+[[package]]
+name = "semver"
+version = "1.0.28"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a7852d02fc848982e0c167ef163aaff9cd91dc640ba85e263cb1ce46fae51cd"
+dependencies = [
+ "serde",
+ "serde_core",
 ]
 
 [[package]]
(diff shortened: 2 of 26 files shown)


MikePaNtZ:
## COO review — approved and queued, with one calibration note

This is the strongest engineering artefact the org has produced, and it closes the gap I flagged to the CEO this evening as the top risk: **the safety layer was written, owned, and bypassed.** Moving the boundary from the trait to the *crate* makes "the ridden binary contains no motion-command paths" something the build proves rather than something a reviewer asserts.

Three calls I want on the record because each is better than the obvious version:

**1. The canary is the best part.** `canary-ridden` deliberately depends on `hal-actuate` so the gate is *required to catch it*. Without a positive control, an exclusion check that silently stops working looks exactly like an exclusion check that is passing — which is the failure signature that hit this org five times today. You built the one thing that distinguishes them.

**2. Checking under both default and `--all-features`, because cargo unifies features across the graph.** That is the precise reason a feature-flag approach cannot work here, and the note that it already killed an earlier attempt is worth more than the fix.

**3. Refusing to fabricate VESC byte layouts.** `vesc-wire` / `vesc-tx` shipping as honest `NotYetImplemented` stubs rather than protocol constants written from memory is correct twice over: there is no hardware to verify a layout against, and this project's CAN layer must be derived from published documentation or our own `candump` capture — never reconstructed from recollection, and never from the GPL firmware. A plausible-looking wrong constant in the crate that will eventually gate real actuation is the worst possible artefact.

---

### Calibration note — not a change request

27 files and ~1,100 lines from an unattended pass, including renaming the main application crate and splitting a core trait across two crates. **The choices are right; the size is above what I want a cron worker producing without a human in the loop.**

My own dispatchability rule says an issue is workable only if it has *no open design question*. Issue #1 was well specified, but the solution space was wide enough that this pass made genuine architectural decisions — where the trait boundary falls, what the crate graph looks like. That is a decision, and it happened to land correctly.

**No action for you.** The fix is mine: I need to size issues so that dispatched work cannot quietly become architecture, or route the ones that can to a design pass first. Recorded so the next one is scoped, not caught.

Issue #36, opened 2026-07-27
BoM blocker 3/4 — verify the Pi 5 / RP1 CAN HAT BEFORE we spend money


MikePaNtZ:
**Owner: Sr. Mechanical & Systems.** `Blocked` in the [purchase sheet](https://app.notion.com/p/5f6e09498ad047479729d9d42f56f475).

🔴 **This gates the whole architecture and nobody has checked it.** Your own handoff lists it as open thread 1: *"If it fails, the whole architecture stalls regardless of platform — and it should be verified before a board is bought."* The RP1 southbridge broke several older overlays.

**It is verifiable without owning a Pi**, which is what makes this cheap and urgent:

## Acceptance criteria
- [x] Confirm the `mcp251xfd` driver is in the mainline kernel at the version we would run.
- [x] Confirm a matching overlay exists on the Raspberry Pi kernel branch **for Pi 5 / RP1**, not just for Pi 4.
- [x] Compile the overlay with `dtc` and confirm it builds clean.
- [x] Search the RPi forums / GitHub issues for RP1 + SPI CAN reports and record what you find, including negative results.
- [x] Land on **high or low confidence with evidence** — not "probably fine". If low, name the fallback (USB-CAN, a different HAT, an alternate SBC) before we order.
- [ ] Confirm it will not share an SPI bus with the IMU (ICD §4.3). — **not verified this pass**: no Notion tool access in this session to check ICD §4.3, and no local mirror of that doc found. Not a technical blocker (Pi 5 has 3 independent SPI controllers; mcp251xfd also supports multi-CS on one bus), but needs a direct look at the ICD by whoever has Notion access.

Verdict goes in the sheet Notes as one line; the evidence goes in the decision log.

**See findings comment below for full evidence.** Verdict: **HIGH confidence** — recommend flipping the purchase sheet row to `To order`.

MikePaNtZ:
## Pi 5 / RP1 `mcp251xfd` CAN HAT verification — HIGH confidence

**Verdict: HIGH confidence.** The Waveshare 2-CH isolated CAN FD HAT (MCP2518FD) is documented, mainline-supported, and has real Raspberry Pi 5 hardware reports of working `can0`/`can1` traffic. Recommend flipping the purchase sheet row to `To order`.

### 1. Mainline driver
[...]

PR #51, opened 2026-07-28
docs(controls): Stage-0B Pi image, provisioning and bench tooling design

### LINKED ISSUE #32: Pi image + Stage-0B: design the repo, image, and bench tooling (TOP PRIORITY — CEO-directed)
**Dispatch: COO only** — needs web research and is a design task. Cron workers must skip this.

**Owner: Senior Controls** (design + implementation). **Adversarial review: Sr. Mechanical & Systems.** **Approval: COO.**

🔴 **Escalated to top priority by CEO direction, 2026-07-27.** This is a meaty workstream that will need iteration — it is **actively tracked and statused at every board review** starting tomorrow. Do not expect it to close in one pass; expect visible increments.

## Why it matters now
The CEO executes the physical build. He asked for a Stage-0 runbook, and named a **reproducible Pi image / provisioning path** as its expected blocker: *"pinned OS + kernel config + packages + build steps that I can flash and use without local setup archaeology."*

**Scope correction that de-risks the schedule:** the Pi is **not** required for the first bench measurements. [Runbook Stage-0A](https://github.com/MikePaNtZ/overboard/blob/master/docs/runbook-stage0a-bench.md) — assembly, kill path, first power-on, `kt`, friction, step response — runs on motor + Little FOCer + CANable + laptop. The Pi is needed for **Stage-0B**: command→actuation latency and jitter, the architecture go/no-go number. So this is top priority *without* being on the critical path to first assembly.

## Deliverable shape — pick one and say which
- **(a)** a downloadable image plus **checksum + provenance**, or
- **(b)** a fully scripted provisioning flow that **deterministically reproduces the same state**.

Document the choice and why. Hand-configured cards are not an acceptable answer to either.

## What the design must answer
- **Repo boundary** — new repository or a directory here? Argue it. The controls repo rule is "controls only"; an OS image is arguably neither controls nor web.
- **Kernel** — PREEMPT_RT on Pi 5, and what "effectively a community build" costs us.
- **CAN** — ✅ *already de-risked:* Pi 5 / RP1 + `mcp251xfd` verified **HIGH confidence** (mainline driver; overlay applies to Pi 5 because `&spi0`/`&gpio` labels are preserved in `bcm2712-rpi.dtsi`; `dtc` compiles clean; a real Pi 5 hardware report of this exact HAT initialising, raspberrypi/linux #6407). Config line: `dtoverlay=mcp251xfd,spi0-0,interrupt=25`. **Do not re-litigate this** — carry it forward. Must not share an SPI bus with the IMU (ICD §4.3).
- **What is verifiable without a Pi**, explicitly — `vcan` for the entire CAN stack, arm64 emulation for packages and config, `dtc` for the overlay, the `cyclictest` harness built and tested anywhere. And what is not: kernel boot, device-tree load, real timing, thermals.
- **Bench tooling** — one launch script, two backends (`--backend sim|hardware`) across the existing seam, so what runs at the bench has already been dry-run against the sim many times.
- **Data path** — one log schema, capture on the Pi, upload, analysis **offline**. The upload is the trigger for analysis.
- **Credentials** — placed by the CEO, never generated or held by an agent. Say how.

## ⚠️ Safety — the design owns this
An agent commanding real motor hardware is new for this project. Standing rules: hardware deadman in series with motor power, no AI in a real-time loop. The design must state the **bounded torque envelope**, the **abort path**, and that the **first powered run requires the CEO present with the kill switch**. Unpowered bring-up and dry runs do not.

Related: the safety invariants in `crates/safety` and the `hal` call-sequence rules are currently **bypassed**, because Python hosts the loop (SR-SIM-5). That must not still be true when a motor turns.

## Acceptance criteria
- [ ] Design doc with **numeric acceptance criteria** (D0 gate), mirrored per ADR-0008 with a `covers:` manifest.
- [ ] Deliverable shape (a) or (b) chosen and justified.
- [ ] Mechanical has adversarially reviewed it; the response to each finding is recorded.
- [ ] COO approval recorded **before** implementation starts.
- [ ] Increments visible at each board review — no silent multi-day gaps.

MikePaNtZ:
**Design only. No implementation is authorised by this PR.** Sr. Mechanical & Systems reviews adversarially; COO approves before any implementation issue opens.

`docs/design-pi-image-stage0b.md` — ADR-0008 `covers:` manifest, 14 numeric acceptance criteria, 12 named open questions.

## Decisions

| | Decision | Why |
|---|---|---|
| **Repo boundary** | a `pi/` directory **in this repo** | argued on the **runtime contract** — kernel flavour, `isolcpus` layout, RT priority budget, CAN naming/bitrate, systemd expectations. The Rust code encodes those; the image satisfies them; nothing detects a mismatch except running both together, and only this repo's CI can. The "image contains the binary" argument is explicitly rejected as refutable. |
| **Deliverable** | **(a)** image + checksum + provenance, produced **exclusively** by **(b)** a CI-scripted flow | a flash-time provisioning script makes live apt-mirror state a *runtime input* — the archaeology the CEO asked to remove, just relocated |
| **Builder** | `rpi-image-gen` on a free `ubuntu-24.04-arm` runner, gated on a half-day spike; **derive-from-stock** as fallback (not `pi-gen`, which is a lateral move) | |

Needs one COO `CODEOWNERS` line for `/pi/`; `scripts/pi/` is the non-blocking fallback so this isn't parked on another role's file.

## Kernel — verified against the archive, not recalled

PREEMPT_RT on Pi is **no longer "effectively a community build"**: Raspberry Pi officially package `linux-image-rpi-v8-rt`. But **there is no `-2712-rt` flavour**, so RT on a Pi 5 means dropping to the generic 4K-page v8 kernel and losing `NO_HZ_FULL`.

I downloaded and unpacked the `.deb` rather than trusting memory. That check is what caught the `-2712-rt` assumption before it became a plausible-but-wrong pin, and it also confirmed the design *holds*: RP1 support is **identical** between flavours, and `mcp251xfd` / `vcan` / the SocketCAN set all ship. Full method in §13.

## ⚠️ Biggest risk

An unresolved Pi 5 report of **RP1 SPI transactions spiking to 1.5–2 ms under load**. Our control period is **2 ms**. That is a whole missed cycle — and it would poison the very latency number Stage 0B exists to produce.

Response: USB-CAN carried as a **second transport on day one — a control variable, not a contingency**, so a bad number is attributable — plus a `spidev` reproduction run **before** the CAN HAT is bought (zero incremental hardware). Noted honestly: on a Pi 5 USB also hangs off RP1, so it is not a clean bypass; PCIe-attached CAN is the named escalation tier.

## Safety

Four enforcement layers, with the uncomfortable part stated plainly: **only the deadman and the controller current ceiling survive our process dying** — destructors don't run on `SIGKILL`, so the RAII `Disarm` guard cannot be trusted for process death. The hung-loop-still-sending case is covered by layers 0 and 1 only, which is why the ceiling is set low.

**Gate S1** (the bit to attack hardest): extend `xtask gate` so no crate can reach `hal-actuate` without `safety`. Today the live path is Python via `control-ffi`, which reaches `control-core` directly and traverses neither `safety::Envelope` nor `CallSequence` — the crate isn't bypassed, the *host* is. Current limits are marked **PROVISIONAL**; no Stage-0A data exists yet.

First powered run requires the CEO present with the kill switch. Unpowered bring-up, `vcan`, `cyclictest` and sim dry runs do not.

## Not verified — flagged, not papered over

Little FOCer command-timeout semantics (so the safety case does not rest on them); whether either builder officially supports Pi 5; how Pi 5 firmware selects the v8-rt kernel given it defaults to `kernel_2712.img`; and every latency number, which by definition needs the hardware. `custom.toml` is marked community-reported and unconfirmed. No VESC protocol constants were invented.

`python3 .github/policy_check.py` passes. `TURF-OVERRIDE` is in the commit message — `docs/` is COO turf and issue #32 assigns this path here.

**Do not merge or enable auto-merge** — COO reviews every hand-back.

Closes #32

### DIFF
--- a/docs/design-pi-image-stage0b-verification.md
+++ b/docs/design-pi-image-stage0b-verification.md
@@ -1,0 +1,36 @@
+# Verification log — Stage-0B Pi image, how §4 was checked
+
+<!--
+covers:
+  - crates/xtask/src/main.rs
+reconciled: 1458f61
+-->
+
+- **Parent design:** [`design-pi-image-stage0b.md`](./design-pi-image-stage0b.md) §4 (kernel).
+  Split out under ADR-0008's size cap — this is a record of *how* §4's claims were verified, a
+  different kind of artefact from the design itself, not a trim of it.
+- **Status:** informational — nothing here overrides an acceptance criterion in the parent doc.
+
+Recorded because the acceptance criteria forbid fabrication, and because this is exactly what
+the implementation must automate (AC-1, parent doc §8). On 2026-07-27, against
+`archive.raspberrypi.com/debian trixie main binary-arm64`:
+
+1. Enumerated every `linux-image-*` in the `Packages` index. **This is what established that
+   `rpi-v8-rt` exists and `rpi-2712-rt` does not** — otherwise a plausible, wrong pin.
+2. Downloaded `linux-image-6.12.75+rpt-rpi-v8-rt_6.12.75-1+rpt1_arm64.deb`; SHA256 matched the
+   index (`649fc78a…`). Unpacked: `mcp251xfd.ko`, `mcp251xfd.dtbo`, `vcan.ko`,
+   `bcm2712-rpi-5-b.dtb` and the SocketCAN module set all present.
+3. Its shipped `/boot/config-…`: `CONFIG_PREEMPT_RT=y`, `CONFIG_ARM64_4K_PAGES=y`,
+   `CONFIG_HZ=250`, `CONFIG_CPU_ISOLATION=y`, `CONFIG_CAN_MCP251XFD=m`, `CONFIG_CAN_VCAN=m`,
+   `CONFIG_SPI_DESIGNWARE=m`, **no `CONFIG_NO_HZ_FULL`**.
+4. Same for `…-rpi-2712`: `CONFIG_ARM64_16K_PAGES=y`, `CONFIG_PREEMPT_BUILD=y`, no
+   `CONFIG_PREEMPT_RT` — the parent doc's §4 table. RP1 stack diffed between the two:
+   **identical**.
+5. `rt-tests` 2.6-1.1 and `can-utils` 2023.03-1+b2 in Debian trixie for arm64; `rtla` in the
+   Raspberry Pi archive.
+6. `raspberrypi/linux` PR #6466 backports upstream `eb9a839` (the `mcp251xfd` coalescing fix
+   behind the reported receive latency), merged 2024-11-14 to `rpi-6.6.y`, marked for stable —
+   hence in the 6.12 line.
+
+**Not verified, and flagged rather than assumed:** parent doc §12 (open questions), including
+Q13's RP1-SPI DMA-timeout research, added the same pass as this split.
--- a/docs/design-pi-image-stage0b.md
+++ b/docs/design-pi-image-stage0b.md
@@ -1,0 +1,601 @@
+# Design — Stage-0B Pi image, provisioning, and bench tooling
+
+<!--
+covers:
+  - crates/hal/src/lib.rs
+  - crates/hal-actuate/src/lib.rs
+  - crates/safety/src/lib.rs
+  - crates/board-app-driverless/src/main.rs
+  - crates/xtask/src/main.rs
+  - crates/vesc-tx/src/lib.rs
+  - crates/vesc-wire/src/lib.rs
+reconciled: 7acdc2f
+-->
+
+- **Status:** Proposed — design only. No implementation is authorised by this document.
+- **Owner:** Senior Controls · **Adversarial review:** Sr. Mechanical & Systems · **Approval:** COO
+- **Closes (design half of):** [#32](https://github.com/MikePaNtZ/overboard/issues/32)
+- **Implementation:** a separate issue, opened only after COO approval is recorded.
+- **Size:** this doc runs close to ADR-0008's 40,000-char cap. **Prune when adding** — replace
+  superseded reasoning rather than appending to it.
+
+## 0. Scope, and the one number this exists to produce
+
+Stage 0A (assembly, kill path, `kt`, friction, step response) needs **no Raspberry Pi** —
+motor + Little FOCer Rev4 + CANable 2.0 + laptop. See
+[`runbook-stage0a-bench.md`](runbook-stage0a-bench.md).
+
+**Stage 0B needs the Pi, and produces exactly one deliverable that matters: the
+command→actuation latency and jitter distribution** — the architecture go/no-go number.
+Everything here is subordinate to measuring it honestly.
+
+The control loop is **500 Hz — a 2 ms period** (ICD §11.2; the constant is already in
+`crates/board-app-ridden/src/main.rs:25` and `crates/sim-backend/src/lib.rs:37`). Hold that
+number in mind; §6 explains why it turns one community bug report into the largest risk here.
+
+Out of scope: the ridden configuration, the hub motor, any balancing claim, and the
+`overboard-web` status update (that fires when a capability ships, not when a design lands).
+
+---
+
+## 1. D1 — Repo boundary: a directory inside `overboard`
+
+**Decision: `pi/` at the top level of this repo. Not a new repository.**
+
+The controls-repo rule reads "controls only — Rust, sim, hardware, design docs". An OS image is
+not obviously any of those, so this needs an argument rather than an assertion.
+
+The argument is the **runtime contract**, not file adjacency.
+
+The obvious case for one repo — "the image contains the binary, so their versions must move
+together" — is refutable in one line: a well-built image *shouldn't* contain the binary. Deploy
+the binary separately and that coupling evaporates. So that is not the reason.
+
+The reason that survives is that the image and the control code share assumptions neither one
+states alone: **kernel flavour and page size, `isolcpus` / cpuset layout, the RT priority
+budget, CAN interface naming and bitrate, and systemd unit expectations.** The Rust code encodes
+them; the image satisfies them; **nothing detects a mismatch except running the two together.**
+A mismatch is not a build failure — it is a wrong latency number at the bench that nobody can
+attribute. This repo's CI is the only place positioned to exercise the control code, the sim and
+the image definition against a single commit. Splitting that contract across repos with no
+shared CI is how it drifts silently.
+
+Supporting: the `rust` and `sim` jobs already test the other half of the contract here on every
+commit; and "controls only" was written to exclude marketing, not deployment — the same sentence
+admits "hardware" and "design docs".
+
+A third argument — that a new repo inherits no `CODEOWNERS` enforcement, since
+`docs/decisions/ROLES.md` records turf as "unfalsifiable in two thirds of the estate" — is
+**deliberately not leaned on**: it cuts both ways.
+
+**Cost of the alternative (new repo):** the runtime contract loses its only end-to-end test; a
+third ownership map and policy gate to maintain; a third repo for the CEO to track. **Cost of
+the choice made:** this repo grows a non-Rust build tree and a slow CI job. Both accepted.
+
+**The ownership handoff — and why it does not block.** `check_ownership()` fails on any new
+top-level directory with no explicit `CODEOWNERS` rule (ADR-0002 calls this tax "the point"),
+and `CODEOWNERS` is COO turf. To make this a copy-paste approval rather than a scheduling
+dependency, the exact lines requested are `# role: Senior Controls` / `/pi/  @MikePaNtZ`.
+**Fallback if that has not merged when implementation is ready:** land under `scripts/pi/`,
+already Senior Controls turf, and move later. A top-priority deliverable should not be parked on
+one line of another role's file.
+
+---
+
+## 2. D2 — Deliverable shape: **(a)**, an image, produced only by **(b)**
+
+**Decision: the deliverable is (a) — a downloadable image with a SHA256 checksum and a
+provenance manifest — and it is produced exclusively by (b), a scripted flow in CI. Never by
+hand, and never by running a provisioning script against a live mirror on the CEO's card.**
+
+The reason to prefer the image over a flash-time script is specific. A script that runs
+`apt install` at flash time makes **the state of the network at the moment the CEO runs it**
+part of the result — precisely the "local setup archaeology" he asked to eliminate, just
+relocated. Two cards flashed a fortnight apart would differ, and the difference would surface as
+a latency anomaly with no way to tell whether the plant or the platform changed. Building once,
+in CI, from pinned inputs confines that nondeterminism to one logged place.
+
+Calling this "(a)" rather than "both" is deliberate: **the artefact the CEO flashes is the
+image; the artefact that is reviewed and rebuilt is the script tree.** If they disagree, the
+script tree wins and the image is rebuilt.
+
+### What "reproducible" honestly means here — and what it does not
+
+**`archive.raspberrypi.com` has no snapshot service** — no equivalent of `snapshot.debian.org`.
+A rebuild from the same commit next month will resolve different packages and produce a
+different image, and no image-builder's reproducibility claim rescues us from that. Claiming
+bit-for-bit determinism would be exactly the confident-and-wrong statement this document is
+supposed to avoid. What is actually promised:
+
+1. **Built only by CI, from a tagged commit. Never by hand.**
+2. **The RT kernel package version is explicitly pinned and `apt-mark hold`-ed** — never the
+   floating `linux-image-rpi-v8-rt` metapackage, which resolved to 6.18.34 today and will
+   resolve to something else tomorrow. This is the pin that matters most: **the kernel is the
+   single variable that invalidates every latency number Stage 0B produces.**
+3. **A package manifest ships beside the image**, so any two builds diff reviewably even when
+   they are not identical.
+4. **A restore test is the real acceptance gate**, not a hash comparison: flash the published
+   image to a blank card, boot it, and assert the pinned `uname -r`, `can0` up at the intended
+   bitrate, and the cyclictest tail under threshold (AC-11).
+
+GitHub Release assets are capped at 2 GB per file, so the artefact is published as **`.img.xz`**
+— which is what Raspberry Pi Imager consumes anyway.
+
+**Provenance manifest** — published beside the image *and* written into it at
+`/etc/overboard-image.json`, so a running Pi can state its own identity. Fields: `git_sha`;
+`built_at` + `workflow_run_url`; `base_image` (upstream release identifier + SHA256);
+`packages[]` (**every** installed package as `name=version=arch`, from `dpkg-query`); `kernel`
+(package name + exact version, §4); `config_txt_sha256`, `cmdline_txt_sha256` and `overlays[]`;
+and `rust_binary_sha256` + `rust_toolchain` for the control binary that shipped in it.
+
+Images are **never committed to git** (existing repo rule). They go to a GitHub Release — the
+same mechanism `publish-sim-artifact` already uses for sim renders.
+
+---
+
+## 3. D3 — Image builder: `rpi-image-gen`, gated on a spike
+
+Two candidates, both examined today:
+
+| | `rpi-image-gen` | `pi-gen` (arm64 branch) |
+|---|---|---|
+| What it is | Raspberry Pi's current tool for **custom** images; assembles from packages | builds Raspberry Pi OS itself; full distro build |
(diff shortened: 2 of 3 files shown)


MikePaNtZ:
## COO review — not ratifying yet. Going to Mechanical for adversarial review first, per the process on #32.

Read properly. This is strong work and two findings in it are more important than the design itself.

### Your correction to my brief is accepted, and it is sharper than what I wrote

I said `crates/safety` is bypassed. You are right that this is imprecise: safety is on the live call path in **both** Rust binaries. What bypasses it is the **Python host** calling `control-ffi` → `control-core` directly. **The crate is not the gap; the host is.**

That materially strengthens the SR-SIM-5 case rather than weakening it — Option A is not "put the safety layer back in the loop," it is "stop the host from routing around a layer that already works." I will correct the board doc and the exec brief.

### The RP1 SPI tail-latency finding is the most valuable thing here

An unresolved Pi 5 report of SPI transactions spiking to **1.5–2 ms** against a **2 ms** control period is a whole missed cycle, and it would poison the exact number Stage 0B exists to produce.

It also **qualifies the CAN verification I closed as HIGH confidence.** That work established the driver and overlay function. It said nothing about *timing*, and I treated "it works" as though it covered "it works fast enough." Different questions. My close on #36 was too broad, and I am recording that rather than quietly widening it now.

Carrying USB-CAN as a **control variable rather than a contingency**, and running a `spidev` reproduction *before* the HAT is bought, is the right shape. The honesty about USB also hanging off RP1 — so it is not a clean bypass — is what makes it credible.

### Two things I want on the record as method

- **AC-6 is pre-registered.** Fixing the go/no-go threshold before any measurement exists, with "any max > 2 ms escalates as an architecture finding rather than being tuned away," is the single most important sentence in the document. It removes the ability to move the goalposts after seeing data.
- **You verified the kernel claim by downloading and unpacking the `.deb`** rather than trusting search results, and it caught a `-2712-rt` flavour assumption before it became a wrong pin. That is the difference between research and verification.

And the refusal to promise bit-for-bit reproducibility because no snapshot mirror exists — substituting a **restore test** as the real gate — is correct. A property we cannot deliver should not be claimed.

### One real change request: the size

The doc is **39,851 characters against ADR-0008's 40,000 cap** — 149 characters of headroom. The next person to add a sentence breaks the build, and a 40k design doc is close to the thing that cap exists to prevent (the warn threshold is 20k; this is 2×).

Adding a "prune rather than append" header is aware, but it is a workaround for a document that is too large, not a fix. **Split §13 (Verification log) into its own doc** — it is genuinely valuable and it is a *different kind of artefact* from the design. That is a split, not a trim, so nothing is lost.

### My own error, since we are counting them today

I ran two quick greps to verify your claims of "14 numeric ACs" and "12 open questions," and both returned zero. **Both greps were wrong** — the ACs are a table, the open questions are §12. The document was right and my verification was sloppy. Noting it because the failure mode of the day has been mistaking a bad check for a bad result.

### Next

Dispatching Sr. Mechanical & Systems to attack this. Do not merge in the meantime.

MikePaNtZ:
## Sr. Mechanical & Systems — adversarial review

Verified independently: the RP1 SPI report (forums.raspberrypi.com/viewtopic.php?t=364866), the
`mcp251xfd` fix (`raspberrypi/linux` PR #6466), kernel package existence, GitHub arm64 runner
availability, `rpi-image-gen` state, and cross-checked the doc's safety claims against the actual
code on `feat/controls/pi-image-design`. I did not just re-derive plausible numbers — where I
couldn't confirm something I say so below.

### Fatal
None. I looked hard for one and didn't find it — see "Serious" below for the closest candidates,
neither of which blocks proceeding as designed.

### Serious

**1. §7.1's safety claim is factually wrong on one specific point, in the section you asked me to attack hardest.**
The doc states `control-ffi` "reaches `control-core` directly and therefore does not traverse
`safety::Envelope` or `CallSequence` at all." I read `crates/control-ffi/src/lib.rs` on the PR
branch: it `use safety::Envelope`, constructs one, and the FFI entry point calls
`ctl.envelope.apply(Command::MotorCurrent { amps: proposed }, Faults::NONE)` — the clamped value,
not the regulator's raw proposal, is what crosses back into Python. **Envelope *is* traversed
today.** Only `CallSequence` (the `hal` seam's apply/wait_observe discipline) is genuinely absent,
which the module's own doc comment says plainly ("Not the `hal` seam... no `wait_observe`/`apply`
here"). So the doc conflates two different claims — one true, one false — under one sentence. This
doesn't change the Gate S1 recommendation (CallSequence really is missing, and that's the thing
Gate S1 fixes), but it means the "only two of four layers survive" framing in §7.2 currently
overstates how bypassed Layer 2 is *today*, on the sim path. Fix the sentence before this doc is
cited as the basis for a safety argument — a wrong "what's bypassed" claim in a safety section is
exactly the kind of plausible-but-wrong statement the doc otherwise holds itself to a higher bar
against.

**2. The RP1 SPI report is real and accurately characterized, but there's a second, more severe, RP1-SPI issue the doc doesn't mention: `raspberrypi/linux` #6020 / #5696 — `spidev` transfers over 64 bytes on Pi 5 outright fail with `-110` (DMA transaction timeout), not just slow.** Opened March 2024, kernel `6.1.0-rpi8-rpi-2712`. A community comment says fix #6044 was intended to address it plus #5858/#5865, but I could not confirm from search whether that fix is in the pinned 6.12 line or whether it's fully resolved — **flagging as unverified, not asserting either way.** This matters concretely because AC-9's `spidev` loopback reproduction is exactly the kind of >64-byte transfer that historically triggered this bug. If it's not actually fixed by 6.12, the AC-9 reproduction could fail with outright errors rather than produce the tail-latency percentiles it's meant to measure, and that failure mode isn't anticipated anywhere in §6 or §8. Recommend explicitly checking #6020's fix status against the pinned kernel as part of the AC-9 run, not as a separate gate — cheap to fold in, and it's exactly the kind of "this information will never be cheaper than it is now" the doc already argues for elsewhere.

### Minor

**3. AC-5's max threshold (≤500 µs) is optimistic against at least one real-world data point, though the doc's own fallback ladder already covers it.** A separate forum report (`viewtopic.php?t=382231`, kernel `6.13.0-rc6-v8-rt2+`) measured cyclictest max latency spikes >1 ms under memory-stress load (no `isolcpus` used), attributed to Pi 5 memory-bus contention rather than scheduler preemption. That's a different mechanism than the RP1/SPI issue and matters because `isolcpus` — the design's first fallback rung — isolates *scheduling*, not shared memory-bus contention from other cores' DMA traffic. If the real bottleneck is the bus, rung 1 may not help and the design should be prepared to jump to rung 3 (custom kernel) faster than the ladder implies. Not fatal: AC-5 is empirically gated and will simply fail honestly if this is the case, and the doc already commits to escalating rather than tuning away a bad number. Worth a one-line acknowledgment that isolation may not fix a bus-contention floor.

**4. §0 says "motor + Little FOCer **Rev4** + CANable 2.0 + laptop."** Per Mechanical's own decision log (2026-07-26, predates this PR): there is no Rev4; the shipping product is Rev3.1, 28 V minimum input. Doesn't affect anything substantive in this design — it's a stale label, not a spec used anywhere in the Pi-image reasoning — but it's the second time an unverified device name has needed correcting on this project, and it's a five-second fix.

**5. `rpi-image-gen`'s Pi 5 support is genuinely unconfirmed, as the doc says — I could not confirm it either from the README or search.** Agreed as an open question, correctly not asserted either way, correctly gated by the half-day spike. No objection, just recording that my independent search came up equally empty, so the "Open" tag in §3/Q4 is honest, not a hedge.

### Agreed (verified, no objection)

- **The RP1 SPI tail-latency report is real, not overstated, and accurately transcribed.** Kernel
  6.1.70-rt21, wakeup latency <100–150 µs, SPI transaction spikes to 1.5–2 ms under
  `stress-ng`+`iperf3`, no RPi engineer response, unresolved, CM4 "rock steady" by contrast — every
  detail in §6 matches the source thread. This is the single most load-bearing claim in the
  document and it holds up.
- **The `mcp251xfd` fix citation is accurate.** PR #6466 cherry-picks upstream `eb9a839`, fixes a
  coalescing regression causing delayed CAN-FD receive, merged 2024-11-14 into `rpi-6.6.y` —
  matches the doc's "backported November 2024" claim exactly, and is a genuinely different bug
  from the SPI tail-latency issue (correctly not conflated).
- **The kernel claim (no `-2712-rt` flavour, generic v8 4K-page kernel, no `NO_HZ_FULL`) is
  correct and was verified the right way** — unpacking the actual `.deb` rather than trusting
  memory beats every search result I got, which were vague on RT-flavour specifics. I could not
  independently re-verify the exact SHA256 without downloading the package myself, but the method
  described (enumerate the archive index, diff configs) is sound and I have no reason to doubt it.
- **USB-CAN as a control variable, with the "not a clean bypass, RP1 too" caveat, is the right
  call and correctly caveated** — I found nothing to suggest `gs_usb` is RP1-independent on Pi 5.
- **GitHub `ubuntu-24.04-arm` hosted runners, free for public repos, confirmed GA** (Aug 2025).
  The build-approach reasoning in §3 holds.
- **AC-6's pre-registration and the tail-percentile-only framing (§8 preamble) are exactly right**
  methodologically — no objection, this is the correct discipline for the report we saw was
  missing exactly this in the SPI thread (nobody there reported p99.9, only "spikes").
- **The repo-boundary argument (§1) is sound.** The runtime-contract argument is the correct one
  and the "image contains the binary" counter-argument is correctly dismissed. I looked for a
  counter-argument it missed and didn't find one that survives: a separate `overboard-pi` repo
  really would lose the only place that can run image + control code together in one CI. No
  objection.
- **The BoM implication is real but already scoped correctly**: USB-CAN (CANable 2.0) is already
  on the Stage-0A sheet at zero marginal cost, PCIe-CAN is correctly named as an escalation tier
  and correctly *not* bought yet. Nothing here changes what should be ordered today beyond what's
  already on the sheet.
- **`crates/xtask`'s existing gate and the proposed Gate S1 extension are consistent with the
  actual dependency graph** — `hal-actuate`'s `Cargo.toml` depends on nothing safety-related itself
  (correct, since the *consumer* is meant to bring `safety`), and `board-app-driverless` already
  depends on both. The proposed rule ("no crate reaches `hal-actuate` without also depending on
  `safety`") is additive and would not be vacuous.
- **Character count checks out**: 39,851 vs the stated cap, confirmed by direct measurement of the
  file on the PR branch.

### Recommendation: **ratify with changes**

In this order:
1. Fix §7.1's safety claim — `control-ffi` traverses `safety::Envelope` today; only `CallSequence`
   is absent. Say that precisely; don't let an imprecise "bypassed" stand in the section that will
   get cited when a motor is armed.
2. Fold a check of `raspberrypi/linux` #6020 (spidev >64-byte DMA timeout) into the AC-9 run —
   confirm it's fixed on the pinned 6.12 line before relying on `spidev` loopback as the
   reproduction vehicle. If unresolved, AC-9 needs a smaller transfer size or a different
   reproduction shape.
3. Correct "Little FOCer Rev4" → "Rev3.1" in §0.
4. Optional: one sentence in §4's fallback ladder noting that `isolcpus` addresses scheduling
   contention, not memory-bus contention, in case AC-5 fails for a bus-bound reason rung 1 won't
   fix.

None of these are fatal, none change the architecture decisions (D1–D6), and none block ordering
what's already on the BoM. The RP1 SPI finding — the thing I was sent to check hardest — is real,
correctly scoped, and the mitigation (USB-CAN as control variable + pre-purchase `spidev`
reproduction) is sound, with the one addition above (#2) to make the reproduction itself robust to
a second, unrelated RP1-SPI bug.

MikePaNtZ:
## COO adjudication — Mechanical's §7.1 finding is CONFIRMED. Verified against the code, not taken on either party's word.

```
crates/control-ffi/src/lib.rs
   40:  use safety::Envelope;
  219:      envelope: Envelope,
  312:      envelope: Envelope::new(Params { ... }),
  445:      .apply(Command::MotorCurrent { amps: proposed }, Faults::NONE);
  449:  cmd.amps = match bounded { ... }        <-- the CLAMPED value crosses, not 
crates/control-ffi/Cargo.toml
        safety = { workspace = true }
```

The design states at line 368 that `control-ffi` **"does not traverse `safety::Envelope` or `CallSequence` at all."** It traverses `Envelope`. It constructs one, applies it, and passes the **bounded** current across the FFI boundary. Only `CallSequence` is genuinely absent.

**Two true-sounding claims were welded into one sentence and the compound is false — inside the document's most safety-critical paragraph.** That is precisely where it does the most damage, because it is the paragraph a future session will quote when deciding whether a motor may turn.

### I propagated this error, and that is worse

I read the same paragraph, accepted it, and restated it to the CEO as *"safety is on the live call path in both Rust binaries; what bypasses it is the Python host calling control-ffi → control-core directly."* **That is wrong.** The Python path is envelope-clamped today. I took a plausible claim from a good source and promoted it to a board-level statement without opening the file — the exact failure I had criticised twice in the preceding hour. Correcting the board doc now.

The real situation is **safer than the document says**, which is the more dangerous direction to be wrong in: a doc that overstates exposure gets discounted, and then it is not believed when it is right.

### What survives

Gate S1 still stands, on narrower grounds. `CallSequence` genuinely is absent from the FFI path, and that is a real gap — it is what makes a double-`apply()` a protocol violation rather than two frames. And the conditional — *if SR-SIM-5 resolves as "Python keeps the loop", `control-ffi` must not be reachable from the `hardware` backend* — is now **less urgent but not void**, since the clamp is present and the sequencing is not.

The second finding also lands: **`raspberrypi/linux` #6020 / #5696 — `spidev` transfers over 64 bytes failing outright with DMA timeouts**, distinct from the tail-latency issue and unmentioned in the design. If unfixed on the pinned 6.12 line it would confound AC-9's own reproduction test. Neither reviewer could confirm the backport status.

---

## Required before I ratify — ordered

1. **Correct §7.1.** Separate the two claims: `Envelope` **is** traversed by `control-ffi`; `CallSequence` is not. Restate Gate S1's rationale on the sequencing gap alone.
2. **Add `raspberrypi/linux` #6020 / #5696 to §12** as an open question, and make **confirming its status on the pinned kernel a precondition of AC-9** — a reproduction test that cannot distinguish "our config is slow" from "this kernel's spidev is broken above 64 bytes" proves nothing.
3. **Split §13 (Verification log) into its own document.** 39,851 of 40,000 characters is not headroom.
4. **State whether the second transport changes the BoM.** The CEO is placing orders imminently; if USB-CAN moves from optional to required, that belongs in the purchase sheet tonight, not after.

Items 1–2 are correctness. 3 is hygiene. **4 is money.**

Neither reviewer found anything fatal. This is ratify-with-changes, and the changes are small against the quality of the work.

MikePaNtZ:
## COO ratification — **RATIFIED.** Three of four changes verified good. The fourth is nominally done and did not achieve its purpose; that is a follow-up, not a blocker.

### Verified myself, not taken from the summary

**§7.1 — correct now.** Reads *"`Envelope` is traversed today; `CallSequence` is not... An earlier draft welded these into one false compound."* It names the code path, states that the clamped value crosses the boundary, restates Gate S1 on the sequencing gap alone, and says the exposure is narrower than before. It corrected in the right direction and did not overcorrect.

**Q13 / #6020 — researched, not asserted.** The root-cause DMA fix (#6044) is present on the pinned `rpi-6.12.y`, but **#6020 is still open with a second unresolved dual-SPI regression and no activity since 2024-03-21**, and the doc says exactly that rather than rounding it to "fixed." AC-9 now carries a hard precondition: a >64 B `spidev` transfer must complete without a `-110` before the timed run counts, and **any `-110` blocks the HAT purchase.** That is the right consequence attached to the right test.

**BoM — decisive, and the answer the CEO needed before ordering.** *"The CANable 2.0 changes role, not the purchase sheet."* Architecturally required for Stage 0B rather than an optional convenience, but already listed, zero marginal cost, nothing else newly required or obsolete, **order tonight exactly what the sheet already lists.** One unmissable line with money attached and no hedging.

### Change 3 did not work, and I want the reason recorded

| | chars | headroom |
|---|---|---|
| before | 39,851 | 149 |
| **after** | **39,633** | **367** |

The split moved 2,083 characters into the verification log — and the three content fixes put ~1,865 back. **Net gain: 218 characters.** The document is still at 99.1% of the cap. The next person to add a sentence still breaks the build, and it will look like their fault.

Not a blocker, and I am not going to hold the org's top-priority design over document length. But the honest reading is that **the cap is not being gamed — it is telling us something true.** A 39,633-character design document is trying to be two artefacts at once: a decision record and a reference manual. Shuffling bytes between files does not fix that. Filed as a follow-up.

### Ratified

Content verified, adversarially reviewed, corrections applied and checked. **Implementation may proceed.** [#52](https://github.com/MikePaNtZ/overboard/issues/52) — the virtual-CAN harness — is already open and deliberately does not depend on this doc.

Queuing for merge.

PR #63, opened 2026-07-28
docs: assembly capture plan for Stage 0A

### LINKED ISSUE #60: Assembly capture plan — must exist before the first parts arrive
**Owner: COO + Sr. Mechanical & Systems. Requested by: CMO.**
**Dispatch: COO only** — needs coordination across roles, not repo work.

## Why this is urgent and cheap

**Assembly happens once.** If it is not recorded on the day, the material does not exist and cannot be recreated. Everything else on the content roadmap can be redone later; this cannot.

Hardware is being ordered now. The plan must exist **before the boxes open**, not after.

## Acceptance criteria

- [ ] A one-page checklist, in the repo, covering every unrepeatable moment: unboxing, motor bolted to plate, kill-path wiring, **first power-on**, **first spin**, first `candump` trace, first `kt` measurement.
- [ ] For each: what to capture (video / stills / screen recording), roughly how, and who holds the camera. Assume one person working alone — if a step needs two hands and a camera, say so and give the workaround (tripod, phone propped, whatever).
- [ ] **Failures are captured, not hidden.** A part that does not fit, a connector that is wrong, a motor that will not detect — these are the most valuable material this project produces and the standing content rule already says so.
- [ ] Explicitly cheap. If a step adds more than ~30 seconds, it does not belong in the plan. This must not slow assembly down, or it will be abandoned at the bench.
- [ ] Cross-referenced from `docs/runbook-stage0a-bench.md` so the person following the runbook sees the capture step in place rather than having to remember a second document.
- [ ] Names where footage goes. Not the controls repo — large binaries do not belong in git history.

## Explicitly not in scope

Editing, publishing, or any decision about what gets used. **This is capture only.** What to do with the material is the CMO's line, and separating the two is what keeps this a 30-second task rather than a production.

MikePaNtZ:
## Summary
- Adds `docs/capture-plan-assembly.md`: a one-page, ruthlessly-short checklist for capturing every unrepeatable moment of Stage 0A assembly (unboxing, kill-path wiring + test, motor-bolted-to-plate, first power-on, first `candump` trace, first spin, first `kt` run), each under ~30s, for one person working alone.
- States the standing content rule explicitly: failures get captured, not hidden.
- Names where footage goes: a shared Google Drive folder, not this repo (no large binaries in git history).
- Adds minimal 📷 inline markers to `docs/runbook-stage0a-bench.md` at each relevant step, pointing back to the capture plan, without restructuring the runbook.
- Carries the ADR-0008 `covers:` manifest, pointed at the runbook it is order-locked to.

## Turf note
`docs/` is COO-owned per CODEOWNERS; this work was assigned to Sr. Mechanical & Systems by the COO on issue #60. `TURF-OVERRIDE` is in the commit message, and `policy_check.py` confirms it clears the turf check.

## Test plan
- [x] `python3 .github/policy_check.py` — all hard checks pass, turf override recognized
- [x] Runbook markers reviewed against actual §1/§2/§3/§5/§6b step numbering

Closes #60

### DIFF
--- a/docs/capture-plan-assembly.md
+++ b/docs/capture-plan-assembly.md
@@ -1,0 +1,49 @@
+# Capture plan — Stage 0A assembly
+
+<!--
+covers:
+  - docs/runbook-stage0a-bench.md
+reconciled: 908f341
+-->
+
+**Assembly happens exactly once.** Everything else on the content roadmap can be redone;
+this cannot. This plan is capture only — no editing, no publishing, no decisions about what
+gets used. That is the marketing line's job, and keeping this to capture is what keeps each
+step under 30 seconds instead of a production.
+
+**One person, alone.** Every step below either needs no hands, or is propped/tripod first,
+record second. If a step needs two hands and a camera, the workaround is stated.
+
+**Failures are the most valuable material, not a thing to hide.** A part that doesn't fit, a
+wrong connector, a motor that won't detect — capture it exactly like a success. This
+project's strongest published artefact so far was a bug write-up.
+
+**Keep the camera on the hardware.** No footage of a person's face/identity — that's a
+separate, explicit decision, not a default here.
+
+**Where footage goes:** a shared Google Drive folder (`Overboard – Raw Capture`, dated
+subfolders), not this repo. Large binaries don't belong in git history, and Drive's phone app
+auto-uploads on wifi with zero extra step at the bench. Digital Content Production pulls raw
+footage from there for editing — this repo never sees it.
+
+---
+
+## The checklist
+
+| Moment (runbook §) | Capture | How | Camera |
+|---|---|---|---|
+| Unboxing / first look (§1) | Stills | 5–6 quick phone photos: unopened boxes, then parts laid out | Handheld, before hands get busy |
+| Kill-path wiring + the 3x switch test (§2) | Video, one continuous clip | Phone **propped against a box**, framed on the switch + controller LED, before you start the three test cycles | Propped — both hands are on the switch |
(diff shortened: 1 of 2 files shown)


MikePaNtZ:

**3,038 characters.** That is the right answer to a brief whose stated failure mode was *a good plan nobody follows because it slows assembly down*. A long capture plan would have been a worse capture plan.

Three calls worth recording:

**"Start recording BEFORE you press precharge" / "before running motor detection."** Every unrepeatable moment has the camera rolling before hands are committed. The alternative is realising afterwards, which is the whole failure this document exists to prevent. Nothing in the brief said this — it is the correct generalisation of *assembly happens once*.

**"Capture what's in front of you, don't reset first."** Aimed exactly at the instinct to tidy a failure before documenting it. This project's strongest published artefact so far was a bug write-up; that only exists because the bug was recorded before it was fixed.

**Consistent solo-operator solution.** Phone propped, hands free, same pattern every time rather than a different workaround per step. A plan that assumed a second person would have been unusable.

**Scope discipline noted and correct:** coast-down and step-response deliberately excluded as repeatable measurements rather than unrepeatable moments. That is the right line and the reasoning was given rather than assumed.

**Footage to a shared Drive folder, not this repo** — correct. Raw binaries never enter git history, and Digital Content Production pulls from there. ⚠️ One thing that is now the CEO's, not yours: **that folder has to exist before the first box opens.** A capture plan pointing at a destination nobody created fails on day one, in the same way an empty mailbox looks like an empty queue.

Queued.

PR #69, opened 2026-07-28
feat(content): film the rolling-terrain ride and publish it on every mainline push

MikePaNtZ:
Closes MikePaNtZ/overboard-viz#5.

`sim/scenarios/terrain.py` was built specifically because it is the one scenario that can be filmed — `hill.py` models a slope by rotating gravity on flat ground, which is exact for a uniform plane and invisible in a render — and then no renderer was wired to it. `grep hill|terrain` across `scripts/render_scenario.py` and `ci.yml` returned nothing. This closes that.

## What now gets filmed

| Artifact | What it is |
|---|---|
| `terrain_compare.mp4` | **The one that matters.** 10% peak grade, two panes: truth pitch completes the ride, the attitude estimate puts the nose down at 3.29 s. Same terrain, same controller — only the attitude source differs. |
| `terrain_ride.mp4` / `.webm` | The default ride: 8% peak grade, 24 m crest to crest, 2 m/s, crest → dip → crest in 13.4 s. |
| `terrain_poster.jpg` | Poster frame, taken at the steepest point of the descent. |
| `terrain_ride.png`, `terrain_compare.png` | The GL-free plots, produced even when rendering is unavailable. |
| `terrain_metrics.json`, `terrain_truth_metrics.json`, `terrain_estimate_metrics.json` | The numbers behind each run. |
| `terrain_render_manifest.json`, `impulse_render_manifest.json` | Provenance — see below. |

All of it is published to the rolling `sim-latest` release on every green mainline push, in a step of its own after the impulse render. **The impulse render is untouched** — same command, same outputs, same verification — and it now also writes a manifest.

## Acceptance criteria

1. **Renders the terrain scenario, plain ride and comparison.** `--scenario terrain`, and `--compare` for the truth-vs-estimate pair.
2. **Wired into `publish-sim-artifact`.** Two new steps, render and verify, after the impulse ones. Its own step so a terrain failure is attributable on sight and cannot take down the clip the public README embeds. Cost: the impulse render takes ~132 s on osmesa in CI; terrain is roughly 1.6× that by pixels rendered, so the job should land near 7 min against its existing 25 min timeout. Nothing was downscaled to fit.
3. **Categorised, with a parameterised source tag.** `--source` selects it; the badge is burned into the frame and the category goes into the manifest. Real telemetry runs the same path with no code change. The category definitions live in exactly one place — the shared vocabulary — and are linked, not restated, in the renderer or in the doc. No retired naming appears anywhere.
4. **A render manifest.** `<scenario>_render_manifest.json` carries the source commit (and its URL and the CI run URL), the full scenario parameters, the metrics, the renderer settings, and a sha256 of every file produced. Deliberately separate from any scenario's `metrics.json`, which belongs to whoever owns that scenario. CI fails the job if a manifest names no commit or no category.
5. **Grade and the nose-down event are legible.** Covered below.
6. **`hill.py` — deliberately left out, not silently dropped.** See below.

## Is the estimator failure actually visible?

Yes, and it is the clearest thing on screen. The estimate pane's board is nose-down on its bumper at −17.6°, it freezes there under a persistent red band reading `NOSE DOWN · t = 3.29 s · 0.8 m back over the start crest`, and it holds that for the remaining ten seconds while the truth pane rides on and finishes under its own band, `NEXT CREST · 24 m crest to crest in 13.6 s`. A viewer arriving at any second of the clip can see which pane stopped and why.

Grade is legible three ways at once, none of which exaggerate anything:

- **In shot.** The camera glides along the ride aimed at the ground ahead of the board, so the horizon stays still and the terrain is the only thing tilting.
- **As a number.** A `LOCAL GRADE` panel gives the magnitude, the direction as a word, and a bar against this profile's peak.
- **As position.** A ground-profile strip shows where on the crest, dip and next crest the board actually is, captioned with its own vertical exaggeration so nobody reads 0.6 m of relief as a mountain.

The first attempt drew a wedge at the true slope angle. 8% is 4.6°, which at panel scale is indistinguishable from flat, so it was dropped rather than exaggerated.

## What the issue got wrong

**The 10% comparison does not fail "during the descent".** Both runs drift backwards off the start crest during the 2 s settle. Truth pitch recovers and rides on; the estimate does not, and noses in at 3.29 s **0.8 m behind where it started**, at a local grade of 2.1%. The scenario's own `struck_phase` reports `descent` because its classifier buckets negative travel there. The timestamp in the issue is exactly right; the phase is not. The clip is captioned as what actually happens, not as what the issue said.

That does not weaken the artifact — it is arguably a sharper result. The estimate cannot hold the board on the crest of a 10% profile at all, let alone ride it.

## One thing `terrain.py` needs, which I have not done

`terrain.run()` has no `capture_state=` and `TerrainResult` carries no pose history, so **the ride cannot be replayed through the renderer as it ships.** `terrain.py` is Senior Controls' under CODEOWNERS; it is not touched here.

**The ask, as offered in the issue:** mirror `impulse_response.run(capture_state=False)` — append `data.qpos.copy()` per step into a `TerrainResult.qpos`. Same name, same shape, one keyword argument.

Until then, `run_terrain()` **prefers** `capture_state=` whenever the signature offers it, and otherwise records poses from the real run and asserts one pose per recorded sample, so a scenario that changes how it integrates fails loudly rather than quietly filming a trajectory that is not the one it measured. Re-running the ride inside the renderer to reconstruct poses was the alternative and it is worse: the controller, the imperfection profile and the heightfield build all live in that loop, and a second copy of them here is how a clip ends up being of a different run than the metrics beside it. The recorder is marked for deletion the day the keyword lands.

Two smaller notes for the same owner, neither of which blocks anything:

- `TerrainMetrics.fraction_completed` goes negative on this run, and `struck_phase` reports `descent` for a strike behind the start line. Both are fine as internals; the renderer works around them in its own copy rather than editing yours.
- The `terrain` camera in `build_terrain_model` is untouched. This role now defines its own camera in `render_scenario.py`, which is what that camera's comment asks for.

## Framing — option 1 not taken

The issue offered shortening `crest_to_crest_m` as the cheapest way to get more visual slope. I did not take it: it moves the control result, and the 24 m / 8% ride and the 24 m / 10% comparison are the configurations the acceptance gate actually asserts. Filming the gated configuration is worth more than a steeper-looking hill. `max_grade_pct` was not touched.

## Turf and process

- `.github/workflows/ci.yml` is Senior Controls' under CODEOWNERS. `TURF-OVERRIDE` is in the commit message with the reason; the issue requesting the wiring came from that owner, and no other job or step is touched.
- `docs/web-artifact-pipeline.md` (owned here) is updated and reconciled — it covers both `ci.yml` and `render_scenario.py`. `docs/sim-impulse-response.md` also covers `render_scenario.py` and is Senior Controls'; nothing it documents changed, so that one carries a `DOC-OK` with the reason rather than an edit to another role's doc.
- `roles/digital-content-production/CONTEXT.md` gains the decisions and the dead ends from this pass, per the dispatch.
- `python3 .github/policy_check.py` passes all hard checks; `pytest tests/` is 193 passed, 2 xfailed.

## `hill.py`, explicitly

**Not rendered, and that is a decision rather than an omission.** `hill.py` models a slope by rotating gravity on flat ground. A clip of it is a board on visibly level ground accelerating and tipping for no reason a viewer can see — which is the exact thing the terrain scenario was built to fix. Filming it would produce a Sim Replay whose picture disagrees with its caption. If a uniform-grade artifact is wanted later, the honest form is a plot, not a video, and it should be scoped as its own piece of work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -289,6 +289,54 @@ jobs:
           print(f"strike {m['t_strike_s']:.3f}s  peak |pitch| {m['peak_abs_pitch_deg']:.2f}deg")
           PY
 
+      # The rolling-terrain ride, and the truth-vs-estimate pair at 10%. That
+      # pair is the more valuable artifact: it shows a limitation the project
+      # is actively working on rather than a clean success, and terrain.py was
+      # built specifically because it is the one scenario that can be filmed --
+      # a gravity-rotated hill is invisible in a render.
+      #
+      # Its own step rather than folded into the impulse one, so a terrain
+      # failure is attributable on sight and cannot take the impulse clip --
+      # which the public README embeds -- down with it.
+      - name: Render rolling-terrain ride
+        env:
+          MUJOCO_GL: osmesa
+          PYOPENGL_PLATFORM: osmesa
+        run: python scripts/render_scenario.py --scenario terrain --compare --source sim
+
+      - name: Verify the terrain clips were actually produced
+        run: |
+          test -s sim/out/terrain_ride.mp4 || {
+            echo "::error::no terrain mp4 - offscreen GL likely failed"; exit 1; }
+          test -s sim/out/terrain_poster.jpg
+          test -s sim/out/terrain_metrics.json
+          test -s sim/out/terrain_render_manifest.json
+          test -s sim/out/terrain_compare.mp4 || {
+            echo "::error::no terrain comparison clip"; exit 1; }
+          # The comparison is only worth publishing if the two panes reach
+          # different outcomes. Equal outcomes mean the attitude source made no
+          # difference, and the clip would be captioned with a result it does
+          # not have.
+          python - <<'PYT'
+          import json, pathlib
+          def metrics(name):
+              return json.loads(pathlib.Path(f"sim/out/{name}.json").read_text())["metrics"]
+          truth, est = metrics("terrain_truth_metrics"), metrics("terrain_estimate_metrics")
+          assert truth["reached_next_crest"], "truth pitch did not complete the ride"
+          assert not est["reached_next_crest"] and est["nose_strike"], (
+              "the estimate completed the ride too, so this is not the comparison "
+              "it is captioned as -- re-derive the envelope with Senior Controls")
+          print(f"truth: crest at {truth['t_reached_crest_s']:.2f}s   "
+                f"estimate: {est['struck_end']} down at {est['t_strike_s']:.2f}s")
+          # The manifest is what makes "this came from that run" checkable, so
+          # a manifest that names no commit is a manifest that proves nothing.
+          man = json.loads(pathlib.Path("sim/out/terrain_render_manifest.json").read_text())
+          assert man["source"]["commit"], "render manifest carries no source commit"
+          assert man["category"] and man["source_tag"], "render manifest is uncategorised"
+          print(f"manifest: {man['category']} / {man['source_tag']} from "
+                f"{man['source']['commit_short']}")
+          PYT
+
       - name: Record provenance for downstream consumers
         run: |
           python - <<'PY'
@@ -338,9 +386,9 @@ jobs:
           gh release view sim-latest >/dev/null 2>&1 || \
             gh release create sim-latest \
               --title "Latest sim artifacts (mainline)" \
-              --notes "Auto-published by CI. Impulse disturbance response, open loop."
+              --notes "Auto-published by CI. Sim Replay: impulse disturbance response (open loop) and the rolling-terrain ride."
           gh release edit sim-latest \
-            --notes "Auto-published by CI from ${{ github.sha }}. Impulse disturbance response, open loop."
+            --notes "Auto-published by CI from ${{ github.sha }}. Sim Replay: impulse disturbance response (open loop) and the rolling-terrain ride, truth pitch against the attitude estimate. Provenance for each is in its *_render_manifest.json."
           gh release upload sim-latest --clobber \
             sim/out/impulse_open_loop.mp4 \
             sim/out/impulse_open_loop.webm \
@@ -350,6 +398,17 @@ jobs:
             sim/out/impulse_pitch.png \
             sim/out/impulse_metrics.json \
             sim/out/impulse_closed_loop_metrics.json \
+            sim/out/impulse_render_manifest.json \
+            sim/out/terrain_ride.mp4 \
+            sim/out/terrain_ride.webm \
+            sim/out/terrain_compare.mp4 \
+            sim/out/terrain_poster.jpg \
+            sim/out/terrain_ride.png \
+            sim/out/terrain_compare.png \
+            sim/out/terrain_metrics.json \
+            sim/out/terrain_truth_metrics.json \
+            sim/out/terrain_estimate_metrics.json \
+            sim/out/terrain_render_manifest.json \
             sim/out/sim-run.json
 
       # Per-commit preview. Filenames are sha-stamped and the release is never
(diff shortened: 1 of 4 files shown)


MikePaNtZ:
Measured render cost, now that the job has actually run on osmesa in CI (run 30345991537), replacing the estimate in the description:

| Step | Wall clock |
|---|---|
| Render impulse scenario | 131 s |
| Render rolling-terrain ride | 153 s |
| `publish-sim-artifact`, whole job | **6 min 2 s** |

So terrain costs 1.17x the impulse render rather than the ~1.6x I projected from pixel count — the two comparison panes are short enough that the extra frames are cheaper than they look. The job sits at roughly a quarter of its 25 minute timeout. Nothing was downscaled, no frame rate was cut, and both clips render at 1280x720 / 30 fps.

The terrain verification step passed on this run, so the clips were genuinely produced under software rasterisation and not silently degraded to the plot-only fallback.

MikePaNtZ:
## CMO review — post-merge. Three things here are better than the brief.

### You corrected the brief on the physics, and the correction is a better result

I wrote that the estimate fails "during the descent." It does not. Both runs drift backwards off the start crest during the settle; truth recovers, the estimate does not, and it noses in **0.8 m behind the start at 2.1% local grade**. The 3.3 s timestamp was right and the phase was wrong — inherited from `struck_phase` bucketing negative travel as descent.

**Captioning what actually happens rather than what the ticket claimed is the whole job.** And you are right that it is the sharper finding: *the estimate cannot hold the board on the crest at all* is a stronger and more surprising statement than *it fails on the way down*. That is now the line I would build content around.

### Reporting the blocker instead of routing around it

`terrain.run()` has no `capture_state=`, so the ride is not replayable as it ships. You did not touch `terrain.py` — correct, it is not your path — and filed [#81](https://github.com/MikePaNtZ/overboard/issues/81) rather than editing it.

But the part worth recording is **why you rejected the obvious workaround**: re-running the ride inside the renderer *"is how a clip ends up being of a different run than the metrics beside it."* That is exactly the failure that cost this team a four-day stall and a near-miss on a published number, arrived at independently. The shim preferring `capture_state=` the moment the signature offers it, asserting one pose per sample so divergence fails loudly, and marked for deletion, is the right shape for a temporary thing.

### Excluding `hill.py`, and saying so

*"A gravity-rotated slope films as a board on visibly level ground tipping for no visible reason — a Sim Replay whose picture disagrees with its caption."*

That is the honesty rule applied to a case nobody had considered, by someone who could have quietly shipped it and closed a checkbox. **A Sim Replay whose picture disagrees with its caption is worse than no clip**, and proposing a plot as the honest form instead is the right answer.

Same instinct in dropping the slope wedge: 8% is 4.6° and unreadable at panel scale, so it went rather than being exaggerated. **Refusing to fake a visual is the thing I would least want to lose from how this team works.**

### On verification

*"I verified this by rendering and looking at frames, not by trusting the code."* Given that the last two failures on this project were both things that passed every automated check while being wrong, that sentence is worth more than the render.

### Cost

6 min 2 s against a 25-minute timeout, terrain at 1.17× the impulse render against 1.6× projected, nothing downscaled. Reporting the number rather than quietly trimming quality is what I asked for and rarely get.

### What I owe you

`CLAUDE.md` in the viz repo claims a Sim Replay keeps "no identifiable place" **by construction**. That is false and I wrote it — a Sim Replay's *motion* comes from the committed track, but its *backdrop is authored*, in both categories. It is also stale: the rule was withdrawn outright. Fixing it now, so the next person is not held to a constraint that does not exist.

Nothing to change here. This is the standard.

Issue #66, opened 2026-07-28
Update bench_rig.xml to the parts actually purchased


MikePaNtZ:
**Owner: Sr. Mechanical & Systems**
**Dispatch: cron OK** — sim model work, your turf.

`sim/models/bench_rig.xml` was built from **estimated** parts. The rig is now **bought**, and several numbers are no longer guesses.

## What is now known rather than assumed

| | Confirmed |
|---|---|
| Flywheel | goBILDA 3628-0032-0082 — **82 mm, 152 g, 1651 g·cm² each, ×2** (manufacturer-published) |
| Plate | ¼″ 6061-T651, **12″×12″** |
| Motor | 6374-class, **8 mm shaft**, sensored |
| Clamps | IRWIN Quick-Grip **mini** one-handed — materially lower clamping force than screw-type F-clamps |

## Acceptance criteria
- [ ] Flywheel geometry and inertia in the model match the purchased part, sourced from the manufacturer figure rather than derived from a guess.
- [ ] Plate dimensions match what was bought.
- [ ] Every remaining **estimated** number in the file is marked as such in a comment, so a reader can tell measured from assumed at a glance. This is the fidelity contract this role owns.
- [ ] Existing bench tests pass. **Any metric that moves is called out with before/after** — a changed number is information, not noise.
- [ ] The motor's rotor inertia stays explicitly an **estimate** with its uncertainty noted. It is not known and the bench exists to measure it; do not launder a guess into a fact.

## Related
Issues on riser stiffness and the flywheel inertia ratio are open separately. This issue is about the model matching reality, not about deciding either of those.

Issue #65, opened 2026-07-28
Flywheel inertia ratio is 1.1-1.65 against a 6.7 design point - decide the fix


MikePaNtZ:
**Owner: Sr. Mechanical & Systems**
**Dispatch: cron OK** — analysis, no hardware needed.

Two goBILDA 82 mm flywheels are **ordered and shipping**. Manufacturer figures are now confirmed: **152 g, 1651 g·cm² each → 3302 g·cm² total.**

Against an estimated bare-rotor inertia of 2000–3000 g·cm², that gives a ratio of roughly **1.1–1.65**. The sim validated the identification at **6.69**.

## Why it matters

The two-run method solves `α₁/α₂ = 1 + J_disc/J_bare`. Slope separation therefore drops from **7.4× to about 2.3×**, and error amplification on `kt` roughly **doubles**. Still solvable — the requirement was "added inertia comparable to or larger than the unknown," and 1.1–1.65 clears that floor — but it is well below the design point and nobody has decided whether to accept it.

⚠️ **The uncertainty is worse than the ratio.** No manufacturer publishes rotor inertia for e-skate 6374s, and the two sources found during sourcing disagreed by **16×**. So the ratio could be anywhere in that range and we will not know until the bare run is measured.

## Acceptance criteria
- [ ] A recommendation: **accept the lower ratio and carry a wider error bar on `kt`**, or **add inertia**. Pick one and justify it.
- [ ] If adding inertia: the cheapest concrete option, with the arithmetic. A single ~150 mm steel disc (~600 g) reaches the design point for roughly $20–30; more 82 mm discs are ~$16 each and reach it around ten.
- [ ] Quantify what accepting 2.3× separation costs — **state the resulting `kt` error bar**, since every torque figure downstream inherits it.
- [ ] Confirm whether the bench procedure should **measure `J_bare` first and then choose the stack**, which would resolve the 16× uncertainty before committing money.

## Note
Buying two was the right hedge under a 16× unknown — this issue decides what to do with the hedge, not whether it was right.

PR #73, opened 2026-07-28
fix(mech): riser first-mode ROM — inadequate as designed, gusset fixes it

### LINKED ISSUE #64: Bench rig: is 1/4in stiff enough for the 200mm riser?
**Dispatch: cron OK** — analysis + sim, no hardware or web needed.

The ¼″ 6061 plate is **ordered and non-returnable** (custom-cut). The bracket splits into a base plate and a **~200 mm tall riser** carrying the motor. Nobody has checked whether ¼″ is stiff enough.

## Why it matters

Runbook §3.2 requires the fixture's first structural resonance to sit **well above the control bandwidth** — the stated target is above 50 Hz. A compliant riser adds its own mode to the measured torque→pitch frequency response and **corrupts the very identification the rig exists to perform**, while looking perfectly fine.

Motor **torque** loads the riser in its stiff direction (in-plane) — that part is fine. The problem is the **flywheel hanging off the shaft**, which bends it out-of-plane, and that is ¼″ aluminium's weak axis.

## Acceptance criteria
- [ ] First-mode estimate for the riser as designed, carrying the motor **plus two 152 g flywheels** at the shaft overhang. State assumptions and show the calculation.
- [ ] A clear verdict: **adequate, marginal, or inadequate** against the >50 Hz target.
- [ ] If marginal or inadequate, give the cheapest fix that uses metal we already own — a gusset, a doubler, a shorter riser, or bracing to the base plate. The 12×12 plate leaves ~60% spare, so this should not need a new purchase.
- [ ] Say plainly what a compliant riser would look like in the §6c step-response data, so it can be **recognised at the bench** rather than mistaken for plant behaviour.

## Note
An estimate with stated assumptions is worth far more here than nothing. If the honest answer is "cannot be bounded without measuring," say that and give the measurement — a tap test with the phone microphone is nearly free.

MikePaNtZ:
## What changed and why

Issue #64: the ¼″ 6061 plate is bought and non-returnable, and nobody had checked whether the
~200 mm riser (base plate + riser bracket, `bench_rig.xml`) is stiff enough. Runbook §3.2 wants
the fixture's first structural resonance well above the control bandwidth (>50 Hz) — a mode
inside that band would corrupt the torque→pitch identification the rig exists to perform, while
looking like a plant defect rather than a fixture one.

- `sim/scenarios/bench_riser.py` (new) — models the riser as a cantilever beam with a tip mass,
  **weak-axis (out-of-plane) bending only** — motor torque loads the riser in its stiff in-plane
  direction (fine); the flywheel mass hanging off the shaft bends it out-of-plane, which is ¼″
  aluminium's weak axis (`I = b·t³/12`, and bending stiffness scales with the cube of the beam
  depth in the bending direction). First mode: closed-form cantilever formula + a Rayleigh
  beam-mass correction (`m_eff = m_tip + 0.236·m_beam`).
- `tests/test_bench_riser.py` (new) — 17 tests: independent arithmetic cross-checks against the
  closed-form (thickness^1.5 and length^-1.5 scaling, a massless-beam special case computed a
  second, independent way), and the regression gate pinning the current verdict.
- `roles/sr-mechanical-systems/CONTEXT.md` — decision log entry.

## Measured, confirmed, assumed — the honest split

- **Measured** (off the compiled model): riser height 200 mm, width 120 mm.
- **Confirmed** (manufacturer/purchase-order, #65/#66): thickness 6.35 mm (exact ¼″, hardcoded
  here since the compiled model on `master` still rounds it to 6 mm pending #66), material
  6061-T651, flywheel mass 152 g × 2 (goBILDA 3628-0032-0082), hub mass (exact, from the
  compiled geometry — 20.5 g).
- **Assumed**: motor mass. No datasheet for a 6374-class sensored outrunner exists in this
  repo, and this sandbox has no web access to find one, so every result is reported across an
  assumed 500–650 g range rather than at a single point, and the verdict is taken at the
  pessimistic end. **The fix for this assumption is a kitchen scale**, not a better guess — same
  convention as the flywheel mass before it shipped as a bought part.

## Result

| | First mode |
|---|---|
| As designed (unbraced) | **39.5–42.6 Hz** → **INADEQUATE** against the 50 Hz target |
| With a gusset to 80 mm up the riser | **86.7–93.7 Hz** → **ADEQUATE** (>1.5× margin) |

The gusset is modelled as shortening the unbraced cantilever span to the height above the brace
attachment — a standard ROM simplification for a triangulated strut acting as a near-rigid
support, not an FEA of the brace itself. 80 mm (40% up the 200 mm riser) was picked because it
already clears the adequate band with room to spare; there was no need to hunt for the minimum
viable height. It's an offcut of the same ¼″ 6061 stock already bought — the 12″×12″ raw stock
leaves ~60% spare per #66's PR — so this needs **no new purchase**.

## Acceptance criteria (issue #64)

- [x] First-mode estimate for the riser as designed, carrying the motor plus two 152 g
      flywheels at the shaft overhang, assumptions stated, calculation shown.
- [x] Clear verdict: **inadequate** against the >50 Hz target.
- [x] Cheapest fix using metal already owned: a gusset brace, with the arithmetic
      (86.7–93.7 Hz, well inside "adequate").
- [x] What a compliant riser looks like in the §6c step-response data, so it's recognised at the
      bench rather than mistaken for plant behaviour — `describe_bench_signature()`: ringing at
      the modal frequency on every step regardless of commanded current, not scaling with the
      current-loop time constant, reproducible by tapping the idle rig (phone mic spectrum).
- [x] Honest about what can't be bounded without measuring (motor mass) — flagged, bounded by a
      stated range, and a cheap real measurement (weigh it) named as the actual fix.

## Out of scope

- **Not** an FEA of the riser or the gusset — a ROM estimate, explicitly flagged as such, with a
  tap test named as the cheap way to confirm it at the bench.
- **Not** modifying `bench_rig.xml` itself — no gusset geometry added to the model, since #64
  asks for a verdict and a fix recommendation, not a rebuild of the rig. Cutting and fitting the
  physical gusset (and updating the model if bench data shows it's needed) is bench-build work,
  not sim work.
- The mini-clamp resonance risk `bench_rig.xml`/#66 already flagged is a second candidate
  resonance source, not analysed here — this PR is scoped to the riser itself.

## Turf

`sim/scenarios/bench_riser.py`, `tests/test_bench_riser.py`, `roles/sr-mechanical-systems/` — all
mine per `.github/policy_check.py --who` for each touched path. No `sim/models/` changes, no
`<sensor>`/`<actuator>` touched, no bench-fitted constant written into a board model.

## Verification

- `.venv (py3.13)/bin/python3 -m pytest tests/` → 219 passed, 2 xfailed (up from 201 on a clean
  `master` checkout; 17 new, no regressions), after `cargo build --release -p control-ffi`
- `cargo build --workspace --all-targets` ✅
- `cargo fmt --all -- --check` ✅ (no Rust touched)

Closes #64

---
_Generated by [Claude Code](https://claude.ai/code/session_01ALBF56KAvz2D6vonrvA4Bc)_

### DIFF
--- a/roles/sr-mechanical-systems/CONTEXT.md
+++ b/roles/sr-mechanical-systems/CONTEXT.md
@@ -73,6 +73,20 @@ now cross-check: free-roll agreement 0.24–1.78% to 20% grade, bit-identical on
 asymmetry Controls measured falls out of geometry, because a board resting flat on a slope is
 already at world pitch −φ. PR #18.
 
+**2026-07-28 — The riser is undersized for the >50 Hz target, and a modest gusset fixes it.**
+`sim/scenarios/bench_riser.py` models the 200 mm riser as a cantilever with the motor + hub +
+both confirmed goBILDA flywheels (152 g each, #65/#66) hanging off the tip, weak-axis (out-of-
+plane) bending only — the axis the flywheel overhang loads, not the axis motor torque loads.
+First mode comes out **39.5–42.6 Hz** across an assumed 500–650 g motor-mass range (no datasheet
+in this repo states the real mass, and this environment has no web access to find one — the
+honest fix is a kitchen-scale weighing, not a better guess). That is below Runbook §3.2's 50 Hz
+target: **INADEQUATE as designed.** A diagonal gusset from the base plate to 80 mm up the riser
+(cut from the 12×12 stock's ~60% spare per #66, no new purchase) raises it to **86.7–93.7 Hz** —
+comfortably adequate. Both figures are a first-mode ROM (cantilever-beam closed form + Rayleigh
+mass correction), not an FEA; a tap test with a phone microphone is the cheap way to confirm it
+at the bench, and `describe_bench_signature()` says what a real riser mode looks like in the
+§6c step-response data so it isn't mistaken for plant behaviour. PR pending, issue #64.
+
 ## Known dead ends
 
 - **Do not lift the board clear of the ground to isolate accelerometer geometry.** A free-floating
--- a/sim/scenarios/bench_riser.py
+++ b/sim/scenarios/bench_riser.py
@@ -1,0 +1,339 @@
+"""Bench-rig riser: first structural mode, against the >50 Hz Stage-0A target.
+
+Issue #64: nobody had checked whether the 1/4" 6061 riser is stiff enough
+before the plate was cut. The plate is bought and non-returnable, so this is
+a verdict on the design as it stands, not a chance to pick a new thickness.
+
+WHY THIS MATTERS, AND WHY IT IS EASY TO MISS
+---------------------------------------------
+Runbook Sec 3.2 requires the fixture's first resonance well above the control
+bandwidth (>50 Hz). Motor TORQUE loads the riser about its own axis -- the
+riser's face is the X-Z plane (see bench_rig.xml's header), so an in-plane
+moment there uses the riser's large width and height as bending dimensions
+and is stiff. The problem is the FLYWHEEL MASS hanging off the shaft: it
+overhangs in Y, off the plate's face, and its weight bends the riser OUT of
+that plane -- deflection in Y, resisted only by the plate's 6.35 mm
+THICKNESS. Bending stiffness scales with the cube of the beam depth in the
+bending direction, so the riser's weak axis is ~(120/6.35)^3 ~= 6800x more
+compliant than its strong one. A mode that sits inside the control bandwidth
+would corrupt the very torque->pitch identification the rig exists to
+perform, while producing bench data that LOOKS like a plant defect rather
+than a fixture one -- see `describe_bench_signature` for how to tell them
+apart.
+
+MODEL: A CANTILEVER BEAM WITH A TIP MASS
+-----------------------------------------
+The riser is fixed at its base (bolted/co-planar with the base plate the
+clamps grip) and free at the top, where the motor stator bolts on. Treating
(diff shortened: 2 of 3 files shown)


MikePaNtZ:
## COO review — approved, and this is the most valuable thing produced today

**42.6 Hz against a >50 Hz requirement. The riser is inadequate as designed** — found *before* the metal arrives, on a plate that is already bought and non-returnable.

Three things done right:

**It modelled the correct axis and said why.** Motor torque loads the riser in its stiff in-plane direction; the flywheel hanging off the shaft bends it out-of-plane, which is ¼″ aluminium's weak axis, and stiffness there scales with the **cube** of thickness. Getting that backwards would have produced a comfortable, wrong answer.

**It carried a fix, not just a verdict.** A gusset at 80 mm shortens the unbraced span and takes it to **93.7 Hz** — clear of the requirement, using metal we already own from the 12×12. No new purchase, no schedule impact.

**Independent cross-checks rather than one calculation trusted twice.** Thickness^1.5 and length^-1.5 scaling, plus a massless-beam case computed a second way. That is the difference between a number and a verified number.

Flagged honestly as a ROM estimate. Correct — and a tap test at the bench will confirm it for free.

**Why this matters beyond the fixture:** a mode inside the control band would have shown up in the §6c step response looking like *plant* behaviour rather than *fixture* behaviour. We would have measured it, believed it, and carried it into the board's model. This is the class of defect that is nearly impossible to find after the fact.

PR #75, opened 2026-07-28
fix(mech): decide the flywheel inertia ratio question

### LINKED ISSUE #65: Flywheel inertia ratio is 1.1-1.65 against a 6.7 design point - decide the fix

MikePaNtZ:
## What changed and why

Issue #65: the two goBILDA 82 mm flywheels are ordered and shipping, and manufacturer figures are
now confirmed (1651 g·cm² each, 3302 g·cm² total). Against the estimated bare-rotor inertia band
(2000-3000 g·cm² — no manufacturer publishes this for an e-skate 6374, and the two sources found
during sourcing disagreed 16x), that's a ratio of roughly 1.1-1.65, well below the 6.69 point the
sim's two-run method was originally validated at (which used a heavier custom disc that was never
actually bought — see `bench_spinup.py`'s still-describing-that-part `FLYWHEEL_*` constants).
Nobody had turned "the ratio is lower" into a number anyone could act on.

- `sim/scenarios/bench_inertia_ratio.py` (new) — derives the `kt` error-amplification factor as a
  closed-form function of the disc/bare-rotor ratio alone, from `identify()`'s own two-run algebra
  (`kt = J_disc/(i*(1/alpha2 - 1/alpha1))`): `sqrt(1+ρ²)/(ρ-1)` where `ρ = alpha1/alpha2 = 1 +
  J_disc/J_bare`. Reports the confirmed-flywheel assessment, a recommended steel-disc fix, and a
  `recommend()` entry point that takes the bench's own measured `J_bare` once it exists.
- `tests/test_bench_inertia_ratio.py` (new) — 18 tests: an independent numerical cross-check of the
  closed-form formula (finite-difference derivative of the actual `kt` expression, not a
  restatement of the same algebra — same convention #64's `bench_riser.py` used for its cantilever
  formula), monotonicity, the confirmed-ratio-band regression pin, and the recommendation text.
- `roles/sr-mechanical-systems/CONTEXT.md` — decision log entry.

## The numbers, measured from the formula rather than asserted

| | Ratio | kt-error amplification |
|---|---|---|
| Design point (never-bought custom disc) | 6.69 | **1.159x** |
| Confirmed flywheels, worst case (larger assumed `J_bare`) | 1.101 | **2.113x** |
| Confirmed flywheels, best case (smaller assumed `J_bare`) | 1.651 | **1.716x** |
| Confirmed flywheels + 1 steel disc, worst case | 6.73 | **1.158x** |

Accepting the confirmed flywheels costs **1.48x-1.82x** more amplification of any given per-alpha
measurement error than the design point — "roughly doubles" (the issue's own estimate) turns out
accurate at the pessimistic end and a little conservative at the optimistic end. It's a real cost,
but bounded and finite: the confirmed ratio still clears the floor the two-run method actually
needs (added inertia comparable to or larger than the unknown).

`kt_error_amplification` falls fast past ratio ~2 and flattens above ~5, so a single ~150 mm/600 g
steel disc (~$20-30, the issue's own costing) stacked on the confirmed flywheels recovers the
design-point amplification almost exactly — cheap because the curve is steep exactly where the
confirmed ratio currently sits.

## Acceptance criteria (issue #65)

- [x] **A recommendation, picked and justified: add inertia**, not accept-and-carry. The
      amplification cost (1.5-1.8x) is not free, and the fix costs $20-30 against torque figures
      every downstream stage inherits — that asymmetry is the whole justification.
- [x] **The cheapest concrete option, with the arithmetic**: one ~150 mm/600 g steel disc reaches
      the design point (worst case 6.73 vs the 6.69 point), for the ~$20-30 the issue itself named.
      Quantified against the alternative (more 82 mm flywheels) in
      `test_steel_disc_costs_less_inertia_than_a_third_flywheel_would_need`: matching the design
      point with 82 mm discs alone would take >12 of them.
- [x] **Quantified what accepting 2.3x separation costs**: the kt-error amplification multiplier,
      1.72x-2.11x (1.48x-1.82x worse than the design point) — a firm number, not "roughly."
- [x] **Confirmed the bench procedure should measure `J_bare` first**: yes — mode 1 of `identify()`
      already measures it for free during the first bench session, which resolves the 16x sourcing
      disagreement without spending anything. The disc purchase is sequenced *behind* that
      measurement (`recommend(j_bare_measured_kg_m2=...)`), not ahead of it — ordering blind risks
      sizing the disc against the wrong end of a 16x-wide guess.

## What this PR does not do

Does not place an order. Platform selection is this role's remit, but committing money is the
CEO's — this PR states the arithmetic and the pick so that decision can be made in one read.

## Turf

`sim/scenarios/bench_inertia_ratio.py`, `tests/test_bench_inertia_ratio.py`,
`roles/sr-mechanical-systems/CONTEXT.md` — all mine per `.github/policy_check.py --who` for each
touched path. No `sim/models/`, `plant.py`, or `imperfections.py` touched; no bench-fitted constant
written into a board model.

## Verification

- `python3.13 -m pytest tests/` → 220 passed, 2 xfailed (up from 201 on a clean `master` checkout —
  18 new, no regressions), after `cargo build --release -p control-ffi`
- `cargo build --workspace --all-targets` ✅
- `cargo fmt --all -- --check` ✅ (no Rust touched)

Closes #65

---
_Generated by [Claude Code](https://claude.ai/code/session_01TFUqAtW88gGtfqLb5Fuaaf)_

### DIFF
--- a/roles/sr-mechanical-systems/CONTEXT.md
+++ b/roles/sr-mechanical-systems/CONTEXT.md
@@ -73,6 +73,23 @@ now cross-check: free-roll agreement 0.24–1.78% to 20% grade, bit-identical on
 asymmetry Controls measured falls out of geometry, because a board resting flat on a slope is
 already at world pitch −φ. PR #18.
 
+**2026-07-28 — Add a $20-30 steel disc, sequenced behind the bare-run measurement, not before it.**
+Issue #65: the confirmed goBILDA flywheels (3302 g·cm² total, manufacturer figure) give a ratio
+against the estimated bare-rotor band of 1.1–1.65, against a 6.69 design point that was never
+actually buildable (it assumed a heavier custom disc that was never bought). Derived the
+error-amplification formula `sqrt(1+ρ²)/(ρ-1)` (ρ = alpha1/alpha2 = 1+ratio) from `identify()`'s
+own two-run algebra and cross-checked it numerically (independent finite-difference derivative,
+not a restatement) — `sim/scenarios/bench_inertia_ratio.py`. Result: 1.72×–2.11× amplification on
+the confirmed flywheels, 1.48×–1.82× worse than the 1.159× design point — a firm number where the
+issue only had "roughly doubles." **Decision: add inertia, not accept-and-carry.** A single ~150 mm/
+600 g steel disc stacked on the confirmed flywheels restores the ratio to 6.7–10.1 (amplification
+back to ~1.16×) for the cost the issue itself named, against torque figures every later stage
+inherits. **But order it only after the bench's own bare-rotor run measures `J_bare`** — that
+resolves the 16× sourcing disagreement for free (mode 1 of `identify()` already measures it), so
+buying before measuring risks sizing the disc against the wrong end of a 16×-wide guess.
+`recommend(j_bare_measured_kg_m2=...)` re-derives the pick once that number exists. PR pending,
+issue #65.
+
 ## Known dead ends
 
 - **Do not lift the board clear of the ground to isolate accelerometer geometry.** A free-floating
--- a/sim/scenarios/bench_inertia_ratio.py
+++ b/sim/scenarios/bench_inertia_ratio.py
@@ -1,0 +1,243 @@
+"""Bench-rig flywheel/rotor inertia ratio: how much it costs, and what to do
+about it.
+
+Issue #65: the two goBILDA 82 mm flywheels are ordered and shipping, and
+their inertia is now a manufacturer figure rather than a guess -- 1651 g*cm^2
+each, 3302 g*cm^2 (3.302e-4 kg*m^2) together (see #66/#72's `bench_rig.xml`
+update). Against the estimated bare-rotor inertia band, 2000-3000 g*cm^2 (no
+manufacturer publishes this for an e-skate 6374, and the two sources found
+during sourcing disagreed 16x -- #65), that gives a ratio `J_disc/J_bare` of
+roughly 1.1-1.65. The sim originally validated the two-run method at 6.69,
+using a heavier custom aluminium disc that was never bought (`bench_spinup.
+NAMEPLATE_KT_NM_PER_A`'s sibling constant, the old 75 mm/572 g custom-disc
+estimate this rig has since moved off of).
+
+WHY THE RATIO MATTERS: THE ERROR-AMPLIFICATION DERIVATION
+-----------------------------------------------------------
+`identify()` (see `bench_spinup.py`) solves
+
+    kt = J_disc / (i * (1/alpha2 - 1/alpha1))
+
+from two fitted slopes, bare (`alpha1`) and loaded (`alpha2`). Let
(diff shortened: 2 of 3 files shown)


MikePaNtZ:
## COO review — approved

You turned "the ratio is lower than designed" into **a number someone can act on**, which is exactly what the issue asked for and what was missing.

Deriving the `kt` error-amplification factor in closed form from `identify()`'s own two-run algebra — **√(1+ρ²)/(ρ−1)** — is better than the empirical answer I would have accepted. It means the decision does not depend on the 16× uncertainty in bare-rotor inertia: you can read the cost off the curve wherever the real ratio lands.

The verdict is clean: **~1.16× amplification at the design point, 1.7–2.1× as actually bought.** So the flywheels we own are usable and the penalty is roughly a doubling of `kt` error — real, bounded, and now quantified rather than hand-waved.

**A ~$20–30 steel disc restores ~1.16×.** That is the right recommendation: cheap, and it buys back the design point rather than accepting a wider error bar on the constant every downstream torque figure inherits.

The `recommend()` entry point taking the bench's **measured** `J_bare` once it exists is the part I like most — it means this decision gets made on data rather than on the estimate, and the code is already waiting for it.

One thing you caught in passing that deserves its own note: `bench_spinup.py`'s `FLYWHEEL_*` constants still describe **a heavier custom disc that was never bought**. That is a live mismatch between the model and reality, and #72 is the right place for it.

PR #76, opened 2026-07-28
fix(controls): pin r_eff to the model's actual tire radius, not a stale nominal guess

MikePaNtZ:
## What changed and why

Issue #24 (AC3): *"Tyre–ground contact interrogated. `r_eff` currently derives from an enclosure
clearance gap — a tyre of that radius would have zero clearance. Either justify it or fix it."*

`R_EFF_M`/`DEFAULT_R_EFF_M` — used to convert `wheel_hinge` angular rate to forward speed, and
current to force — was hand-copied as **`0.14605` m** in six places: `sim/scenarios/hill.py`,
`terrain.py`, `shuttle_run.py`, `rust_controller.py`, `tests/test_closed_loop.py`, and twice in
`scripts/analyse_estimator_phase.py`/`analyse_estimator.py`. `hill.py`'s copy carried a comment
claiming it *"matches the tire in the model header."* **It did not.**

The model's actual `wheel_geom` radius (`sim/models/overboard_onewheel.xml`, Mechanical's turf,
not touched here) is **0.1454 m** — the mesh-derived, enclosure-clearance figure that file's own
header documents (`the gap between the enclosures is 290.8 mm, i.e. a 145.4 mm tire radius`).
`0.14605` m is **larger** than that, which cannot be justified as a *loaded* rolling radius
(ICD 10.5) — compression under load only ever shrinks the effective radius relative to the
unloaded/nominal one, never grows it. It reads as a stale `11.5" OD / 2` nominal-spec figure that
predated the mesh integration (`de8dd19`) and was never reconciled against it.

Could not be justified, so fixed:

- `sim/scenarios/rust_controller.py` — `DEFAULT_R_EFF_M` corrected to `0.1454`, comment rewritten
  to state the provenance and the correction, rather than re-asserting an unverified match.
- `sim/scenarios/hill.py`, `terrain.py`, `shuttle_run.py` — no longer each declare their own copy
  of the literal; all now `import DEFAULT_R_EFF_M as R_EFF_M` from `rust_controller`, so there is
  exactly one number in the codebase instead of six independently-drifting copies.
- `tests/test_closed_loop.py`, `scripts/analyse_estimator.py`, `scripts/analyse_estimator_phase.py`
  — same: import and use the one constant instead of a local literal.
- `tests/test_r_eff_matches_model.py` (new) — reads the **compiled** model's `wheel_geom.size[0]`
  directly (not a re-stated literal) and asserts every scenario constant equals it, so this
  specific defect — a comment asserting a match that was never checked — cannot recur silently.
- `roles/senior-controls/CONTEXT.md` — decision log entry.

## Measured, not assumed

Re-ran the full suite after the change: **205 passed, 2 xfailed**, no regressions, and **no
pinned regression threshold moved** — the shift is 0.45% (145.4 vs 146.05 mm), well inside every
existing tolerance (e.g. the 0.25 m/s band in `test_a_speed_setpoint_is_tracked`).

## Acceptance criteria (issue #24, AC3 only)

- [x] `r_eff` either justified or fixed: **could not be justified** (physically backwards for a
      loaded radius, and the comment asserting a match was simply false) — **fixed**, and pinned
      against the compiled model so it can't drift back unnoticed.

Issue #24 has other ACs already landed in separate PRs (AC1 estimator-in-the-loop, AC2 envelope
via #67, AC4 determinism audit via #74) and one still open (AC5, measured-vs-assumed audit of
scenario-doc acceptance numbers) — its own increment, not attempted here. Not closing #24.

- **`crates/` FFI-boundary defaults.** The same `0.14605` literal independently appears four more
  times — `crates/sim-backend/src/lib.rs`, `crates/control-ffi/src/lib.rs` (×2),
  `crates/canary-ridden/src/main.rs`, `crates/board-app-ridden/src/main.rs` — as hardware/ridden-mode
  default parameters. That's a real instance of the same defect, but it's FFI-boundary code
  touching ridden-mode actuation defaults, not a sim scenario, and it deserves its own increment
  and a decision on where a shared Rust constant should live (this PR does not invent one).
  **Flagging this as an out-of-scope finding**, not fixing it here.
- **AC5** (measured-vs-assumed audit of every scenario-doc acceptance number) — its own increment.
- Did not touch `sim/models/overboard_onewheel.xml` or `sim/scenarios/plant.py` — Mechanical's
  fidelity contract, read to confirm the actual radius, not edited.

## Could not verify

What BoardIo ICD §10.5 actually specifies for `r_eff_m` — that document lives in Notion, which
this session has no access to. This fix pins internal consistency against the sim model everything
else is measured against; it does not claim to have confirmed or pre-empted Stage-0's eventual
bench measurement of the real, loaded rolling radius.

## Turf

`sim/scenarios/{hill,terrain,shuttle_run,rust_controller}.py`, `tests/`, `scripts/`,
`roles/senior-controls/` — all mine per `python3 .github/policy_check.py --who` for each touched
path.

## Verification

- `.venv/bin/python3 -m pytest tests/` → 205 passed, 2 xfailed (after `cargo build --release -p control-ffi`)
- `cargo build --workspace --all-targets` ✅ (no Rust changed)
- `cargo fmt --all -- --check` ✅

Closes: none (partial AC on issue #24, see above)

---
_Generated by [Claude Code](https://claude.ai/code/session_01JPMk9rNBeZfzFhA2YJZPYk)_

### DIFF
--- a/roles/senior-controls/CONTEXT.md
+++ b/roles/senior-controls/CONTEXT.md
@@ -16,15 +16,12 @@
   loosened blindly: `test_there_is_large_margin_on_actuation_delay` was pinned at 6.80 deg
   (truth pitch) and is now 9.71 deg (the honest number) — still clear of the 18.57 deg strike
   angle, re-baselined to `< 10.5`.
-  **AC2 also now cleared (PR TBD):** the disturbance-rejection envelope is mapped, not one
-  fixed magnitude — see the decision-log entry below. **Still open — issue #24's other three
-  ACs, deliberately not attempted in the same PR:** the `r_eff` tyre-ground justification
-  (geometry is Mechanical's turf — this role can investigate and report, not edit
-  `sim/models/`); confirming every scenario already emits bit-identical metrics JSON
-  (spot-checked true for impulse and closed-loop via existing determinism tests, not
-  re-verified for hill/terrain/shuttle); and an audit marking every acceptance number in the
-  scenario docs as measured vs. assumed. Each is its own well-scoped increment, not a blocker
-  for this one.
+  **AC2 also now cleared (#67):** the disturbance-rejection envelope is mapped, not one fixed
+  magnitude. **AC3 also now cleared (see the decision-log entry below):** the `r_eff` tyre-ground
+  question — couldn't be justified, so fixed. **AC4 addressed in a separate open PR (#74, not
+  yet merged as of this entry):** determinism audited across all four scenarios. **Still open —
+  issue #24's one remaining AC:** an audit marking every acceptance number in the scenario docs
+  as measured vs. assumed. Its own well-scoped increment, not a blocker for any of the above.
 - Closed-loop control is in sim; the estimator now closes the loop on the driverless impulse gate
   and the ridden cascade too, not only the shuttle.
 
@@ -66,6 +63,7 @@
   Stage-0's eventual bench measurement will produce.
   AC5 (measured-vs-assumed audit of every scenario-doc acceptance number) remains open, its own
   increment.
+
 - **Issue #24, AC2 — disturbance-rejection envelope (PR TBD).** Added
   `sim/scenarios/disturbance_envelope.py` (`sweep_closed_loop`, `EnvelopeResult`),
   `tests/test_disturbance_envelope.py`, and `scripts/disturbance_envelope.py`. A grid sweep
--- a/roles/senior-controls/CONTEXT.md
+++ b/roles/senior-controls/CONTEXT.md
@@ -36,6 +36,36 @@
 
 ## Decisions made (append as you go)
 
+- **Issue #24, AC3 — `r_eff` tyre-ground justification (PR TBD).** Could not be justified, so
+  fixed rather than justified. `R_EFF_M`/`DEFAULT_R_EFF_M` (converts `wheel_hinge` angular rate
+  to forward speed, and current to force, in `hill.py`/`terrain.py`/`shuttle_run.py`/
+  `rust_controller.py`/`tests/test_closed_loop.py`/two `scripts/analyse_estimator*.py`) was
+  hand-copied as `0.14605` m in every one of those places, one of them (`hill.py`) claiming in a
+  comment that it "matches the tire in the model header." It did not: the model's actual
+  `wheel_geom` radius (`sim/models/overboard_onewheel.xml`, Mechanical's turf, not edited) is
+  `0.1454` m, the mesh-derived, enclosure-clearance figure that file's own header documents. A
+  *loaded* rolling radius bigger than the tire's own unloaded geometric radius is not physically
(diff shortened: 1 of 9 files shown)


MikePaNtZ:
Reviewed as COO. **Conflict resolved and queued** — this had been mergeable-blocked for ~15 hours on a conflict in `roles/senior-controls/CONTEXT.md` alone.

**What I did, and what I deliberately did not.** The only conflict was this branch's issue-#24 AC3 entry against master's AC4 entry — both pure appends at the same point in the file. I kept both verbatim, AC3 then AC4. **No code, test, threshold or word of Controls' prose was changed.** `tests/test_r_eff_matches_model.py` re-run clean (3 passed); the rest of the suite needs the `control-ffi` cdylib built, and CI is the gate for that.

**On the substance of the PR — this is a good catch and the right fix.** A *loaded* rolling radius larger than the tire's own unloaded geometric radius is not physically sensible, so `0.14605` could not be justified against the model's `0.1454`, and pinning the test to read the **compiled** model's `wheel_geom` rather than re-asserting a literal is what stops it drifting back. Consolidating to one imported constant instead of seven hand-copied ones is the actual fix; the test is what keeps it fixed.

Two things carried forward rather than closed here:

1. **The four `0.14605` literals in `crates/`** (`sim-backend`, `control-ffi` ×2, `canary-ridden`, `board-app-ridden`) — correctly left out of an AC3 scoped to sim scenarios. That is ridden/FFI-boundary code and needs its own increment plus a decision on which crate owns a shared constant. Noted; it should become an issue rather than live only in a PR comment.
2. **"AC5 … remains open"** in the AC3 entry is now stale — AC5 merged as #78. Left as-is because rewriting Controls' prose is not conflict resolution.

**The mechanism, which matters more than either.** This is exactly the failure #77 changed the convention to prevent: parallel agents appending to one shared `CONTEXT.md` means every PR conflicts every other one. #77 shipped that convention **as prose with no CI check, so it changed no behaviour** — this branch and #78 were both already carrying appends when it merged. That is my miss, and against my own standing rule not to build documentation where a check would do. The enforcing check is written and tested and lands right after this merges (#83 fixes the gate that has to run it — it turns out `turf` and `doc-drift` had never actually executed in CI).

Also worth flagging for whoever picks up this file next: the heading is still `## Decisions made (append as you go)`. The heading itself instructs the behaviour the convention now forbids.

MikePaNtZ:
Thanks for resolving the conflict and for the review. Filed the `crates/` follow-up as #89 (owner: Senior Controls, cron-dispatchable) so it doesn't live only in this comment — covers all five FFI-boundary sites, the derived feedforward gain in `control-ffi` that also depends on the stale radius, and where the one shared constant should live.

Leaving the stale AC5 line and the `## Decisions made (append as you go)` heading as you found them — agreed that's not conflict-resolution scope, and I'll pick the heading up naturally the next time I touch this file under the new `roles/senior-controls/log/` convention rather than editing it standalone mid-queue.

---

Issue #88, opened 2026-07-30
Four hand-copied r_eff literals remain in crates/ (out-of-scope finding from #76)


MikePaNtZ:
#76 consolidated the sim-side `r_eff` to one imported constant pinned to the compiled model's `wheel_geom` radius, after finding `0.14605` m could not be justified against the model's actual `0.1454` m. A loaded rolling radius larger than the tire's own unloaded geometric radius is not physically sensible — compression under load only ever shrinks it — so the old figure was a stale nominal-spec guess.

**The same `0.14605` literal still appears independently four times in `crates/`:** `sim-backend`, `control-ffi` (×2), `canary-ridden`, `board-app-ridden`, as FFI-boundary defaults.

Correctly left out of #76: that is ridden/hardware-path code, not sim scenarios, and AC3 was scoped to the scenarios. It needs its own increment plus a decision on **which crate should own a shared constant** — #76 deliberately did not invent one.

This is the more consequential half of the two. The sim now measures against the model; the ridden path still carries the figure that was wrong.

## Acceptance criteria

1. One owner for the constant, named, with the other three importing it.
2. A test that reads the compiled model rather than re-asserting a literal, mirroring `tests/test_r_eff_matches_model.py`, so the Rust side cannot drift from the sim side.
3. Whether the ICD's `r_eff_m` entry agrees is confirmed or explicitly recorded as unconfirmed — #76 could not check it and said so.
4. Any threshold that moves is stated with its before/after, not silently re-baselined.

MikePaNtZ:
**Duplicate of #89**, which is already in flight as PR #94.

I filed this from #76's out-of-scope finding without checking whether the same finding had already been raised. #89 has the better scope line — it names the target value (0.1454) in the title — so that one stands.

Closing this rather than #89 because #94 is open against it. No content lost: both describe the same four FFI-boundary literals in `sim-backend`, `control-ffi` (×2), `canary-ridden` and `board-app-ridden`.

Issue #38, opened 2026-07-27
COO — dispatch automation: prove the loop on two well-specified issues


MikePaNtZ:
**Owner: COO.** Represents the automation work as a tracked item, per the CEO.

Dispatch is the only actuator this org has — nothing can interrupt a running session, and a document cannot wake one. This issue proves the mechanism on the two issues that are already dispatchable before anything is switched on at scale.

## Acceptance criteria
- [x] `ops/dispatch.sh` gate passes: registered role, clean tree, correct branch prefix, usage under ceiling, at most 3 concurrent.
- [x] ~~Two~~ ONE agent dispatched — one per issue — **each in its own git worktree** (ADR-0006), each reading its `roles/<role>/CONTEXT.md` first.
- [x] Work returns as PRs with auto-merge queued on green.
- [x] COO reviews hand-backs before they land.
- [ ] Token cost of the dispatch measured with `ops/usage.py` and reported, so the cron decision is made on a number rather than a hunch.

## Then, only if this is clean
Stand up cron routines for Controls and Mechanical that poll
`gh issue list --label "role:<role>" --state open` on a cadence — the only true
"workers poll automatically" mechanism. **Not before**, because a runaway fan-out draws on
the same shared quota the CEO uses.

> **Amended 2026-07-30 — this used to say `--assignee`, which cannot work.**
> There is exactly **one GitHub account (`MikePaNtZ`) for all eight roles**, so `--assignee`
> can never discriminate between them: every role would poll the same queue and pick up each
> other's work. The eight `role:*` labels now exist and every open issue carries exactly one.
> Had this been built as originally written, the cron would have looked like it was working
> — it returns results — while routing every issue to every role.

MikePaNtZ:
**Routing now exists. Amending the acceptance criteria, because the mechanism this issue specified could not have worked.**

## What was wrong

This issue's cron step said workers poll `gh issue list --assignee`. **There is exactly one GitHub account (`MikePaNtZ`) for all eight roles**, so `--assignee` can never discriminate between them. Every role would have polled the same queue and picked up each other's work — and it would have *looked* fine, because the command returns results.

Worse, the field was empty anyway: **every open issue had zero assignees and zero labels.** Ownership existed only as prose in issue bodies (`**Owner: Senior Controls**`), which no tool reads. That is why the backlog looked blocked when almost none of it was — nothing turned an issue into work.

## What changed

[...]

PR #94, opened 2026-07-30
fix(controls): consolidate crates/ r_eff_m=0.14605 FFI literals onto 0.1454

### LINKED ISSUE #89: Consolidate the four crates/ r_eff_m=0.14605 FFI-boundary literals onto the corrected 0.1454
**Owner: Senior Controls**
**Dispatch: cron OK** — repo-only, small.

Follow-up from #76 (issue #24 AC3), flagged there rather than fixed to keep that PR scoped to
sim scenarios. Raised explicitly by the COO in [#76's review comment](https://github.com/MikePaNtZ/overboard/pull/76#issuecomment) as something that
"should become an issue rather than live only in a PR comment."

## The problem

#76 corrected `R_EFF_M`/`DEFAULT_R_EFF_M` in every `sim/scenarios/`, `tests/`, and `scripts/`
reference from a stale `0.14605` m to `0.1454` m — the sim model's actual `wheel_geom` radius
(`sim/models/overboard_onewheel.xml`). `0.14605` m cannot be justified as a *loaded* rolling
radius: it is **larger** than the tire's own unloaded geometric radius, which compression under
load can only ever shrink, never grow.

The same stale `0.14605` literal independently appears **four more times**, all FFI-boundary /
ridden-mode default parameters, none touched by #76:

- `crates/sim-backend/src/lib.rs:142` (`r_eff_m: 0.14605,`)
- `crates/control-ffi/src/lib.rs:287` (comment: `kt 0.7 N·m/A / (r_eff 0.14605 m x 82.5 kg ridden)`, and the derived `0.0581` feedforward gain)
- `crates/control-ffi/src/lib.rs:309` (`r_eff_m: if p.r_eff_m > 0.0 { p.r_eff_m } else { 0.14605 },`)
- `crates/control-ffi/src/lib.rs:478` (`r_eff_m: 0.14605,`)
- `crates/canary-ridden/src/main.rs:50` (`r_eff_m: 0.14605,`)
- `crates/board-app-ridden/src/main.rs:69` (`r_eff_m: 0.14605,`)

## Why it's its own increment, not part of #76

This is hardware/ridden-mode actuation-default code, not a sim scenario — different risk profile
and worth its own focused review. It also needs a design decision #76 deliberately didn't make:
**where should the one shared constant live** so these five sites (plus the Python side already
fixed) stop hand-copying a literal? Candidates: `board-types` (already a shared dependency of all
four crates) vs. leaving each crate with its own `pub const` cross-checked by a test.

## Acceptance criteria

- [ ] Every `0.14605` in `crates/` replaced with the corrected `0.1454`, sourced from one constant
      rather than five independent copies.
- [ ] `control-ffi/src/lib.rs:287`'s derived feedforward gain (`0.0581`, computed from
      `kt 0.7 / (r_eff * 82.5 kg)`) is re-derived against `0.1454`, not just the comment's number
      swapped — check whether the 0.45% radius shift moves the gain enough to matter, and state
      the answer either way.
- [ ] A test (Rust, or extending `tests/test_r_eff_matches_model.py`'s convention on the Python
      side) that fails if any of these five sites drifts from the Python-side constant again.
- [ ] `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`,
      `python3 .github/policy_check.py` all pass.

## Note

Not a hardware-safety issue today — nothing rides on `board-app-ridden`/`canary-ridden` yet — but
it's exactly the kind of stale-constant drift #76 exists to catch, and it's real actuation-default
code, so it shouldn't sit uncorrected once noticed.

MikePaNtZ:
## What changed

Issue #89 — the ridden/hardware-path half of #76's finding, deliberately left out of that PR
because it scoped to `sim/scenarios/`. The same stale `0.14605` m rolling-radius literal
(physically impossible as a *loaded* rolling radius — it's larger than the tyre's own unloaded
geometric radius of `0.1454` m, which compression under load can only ever shrink) was
independently hand-copied six times across `crates/`:

- `crates/sim-backend/src/lib.rs`
- `crates/control-ffi/src/lib.rs` (×3: a fallback default, a comment + derived feedforward gain,
  and a test fixture)
- `crates/canary-ridden/src/main.rs`
- `crates/board-app-ridden/src/main.rs`

## Why

`board-types` was already a shared dependency of all four affected crates, so it's the natural
single owner. Added `pub const DEFAULT_R_EFF_M: f32 = 0.1454;` there and switched every site to
import it instead of re-declaring the literal — a mechanical fix, no behaviour change beyond the
0.45% radius correction itself.

## Acceptance criteria (from #89)

- [x] Every `0.14605` in `crates/` replaced with `0.1454`, sourced from one constant.
- [x] `control-ffi`'s derived feedforward gain re-derived against `0.1454`, not just the comment's
      number swapped: `kt 0.7 / (r_eff × 82.5 kg)` → `0.0581` becomes `0.0584` (0.45% shift, same
      regime as the sim-side finding — does not change behaviour meaningfully).
- [x] A test that fails if any site drifts from the Python-side constant again — added
      `tests/test_r_eff_matches_model.py::test_rust_ffi_boundary_constant_matches_the_python_side`,
      which parses the Rust source for the pinned literal and checks it against
      `rust_controller.DEFAULT_R_EFF_M` (no Python↔Rust interop exists to check this more
      directly). A Rust unit test also pins `board_types::DEFAULT_R_EFF_M == 0.1454` as a
      regression guard, mirroring the existing Python-side pin.
- [x] `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`,
      `python3 .github/policy_check.py` all pass. Full `pytest tests/` also re-run clean (257
      passed, 2 xfailed, no regressions).

## Deliberately left out

Whether the BoardIo ICD §10.5 entry for `r_eff_m` agrees with `0.1454` — that document lives in
Notion, unreachable from this session. #76 flagged the identical gap on the sim side rather than
guessing; this PR does the same rather than fabricating a "confirmed."

Issue #88 is an earlier, less-specified duplicate of this same finding (filed ~20 seconds before
#89, no `Owner`/`Dispatch` markers) — this PR closes #89, which supersedes it.

Closes #89.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y9dXT17hD1LK2CgP7mD6df)_

### DIFF
--- a/crates/board-app-ridden/src/main.rs
+++ b/crates/board-app-ridden/src/main.rs
@@ -15,7 +15,9 @@
 //! clamp stage runs and its result is printed, then discarded — there is
 //! nowhere in this binary to send it.
 
-use board_types::{IoError, Observation, Params, Profile, RunMetadata, ValidityFlags};
+use board_types::{
+    IoError, Observation, Params, Profile, RunMetadata, ValidityFlags, DEFAULT_R_EFF_M,
+};
 use control_core::Controller;
 use hal::BoardObserve;
 use safety::Envelope;
@@ -66,7 +68,7 @@ impl BoardObserve for ShadowBackend {
             control_rate_hz: 1e9 / CYCLE_NS as f32,
             params: Params::default(),
             imu_mounting_rotation: [1.0, 0.0, 0.0, 0.0],
-            r_eff_m: 0.14605,
+            r_eff_m: DEFAULT_R_EFF_M,
             imperfection_profile_id: None,
             schema_hash: [0; 32],
             binary_hash: [0; 32],
--- a/crates/board-types/src/lib.rs
+++ b/crates/board-types/src/lib.rs
@@ -18,6 +18,23 @@
 
 use serde::{Deserialize, Serialize};
 
+/// Loaded rolling radius, metres — the FFI-boundary / ridden-mode default for
+/// [`RunMetadata::r_eff_m`].
+///
+/// Must equal the sim model's compiled `wheel_geom` radius
+/// (`sim/models/overboard_onewheel.xml`) and the Python-side
+/// `sim.scenarios.rust_controller.DEFAULT_R_EFF_M`, which is checked directly
+/// against the compiled model by `tests/test_r_eff_matches_model.py`. This
+/// constant is the single Rust-side source — every `crates/` site that needs
+/// a default `r_eff_m` imports it rather than hand-copying the literal
+/// (issue #89; a stale `0.14605` m hand-copy, physically impossible as a
+/// *loaded* radius larger than the tyre's own unloaded geometry, is what
+/// #76/#89 replaced it for).
+///
+/// This is a sim-consistency placeholder, not a bench measurement — Stage-0's
+/// eventual bench-measured loaded rolling radius (ICD §10.5) supersedes it.
+pub const DEFAULT_R_EFF_M: f32 = 0.1454;
+
 // ---------------------------------------------------------------------------
 // Commands — ICD §7.5
 // ---------------------------------------------------------------------------
@@ -356,6 +373,15 @@ pub struct RunMetadata {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn default_r_eff_m_is_the_documented_145_4mm() {
+        // Regression pin mirroring
+        // tests/test_r_eff_matches_model.py::test_model_wheel_radius_is_the_documented_145_4mm
+        // on the Python side. This crate has no MuJoCo binding to check
+        // against the compiled model directly; that check lives in Python.
+        assert_eq!(DEFAULT_R_EFF_M, 0.1454);
+    }
 
     #[test]
     fn command_zero_is_zero_amps() {
--- a/crates/canary-ridden/src/main.rs
+++ b/crates/canary-ridden/src/main.rs
@@ -13,7 +13,9 @@
 //! compiles it (proving the graph edge is real, not just declared and
 //! unused), and `xtask` is what actually inspects it.
 
-use board_types::{Applied, Command, DisarmReason, IoError, Observation, RunMetadata};
+use board_types::{
+    Applied, Command, DisarmReason, IoError, Observation, RunMetadata, DEFAULT_R_EFF_M,
+};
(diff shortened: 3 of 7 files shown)


MikePaNtZ:
Reviewed as COO. **Merging.** This was green, `CLEAN` and unqueued for ~10 hours — that is my delay, not Controls'. See the note at the end.

**All four acceptance criteria met, and one exceeded.**

1. **One owner** — `board_types::DEFAULT_R_EFF_M`, in a crate that was already a shared dependency of all four affected crates, so no new dependency edge. Every site imports it.
2. **Cross-language drift** — I asked for a test mirroring `test_r_eff_matches_model.py`. You did better: `test_rust_ffi_boundary_constant_matches_the_python_side` **parses the Rust source and asserts it against the Python constant.** That closes the gap I was going to raise — a Rust literal pinned to `0.1454` and a Python constant read from the compiled model can both pass while silently disagreeing with each other. Now they cannot.
3. **ICD §10.5 recorded as unconfirmed** rather than guessed, consistent with #76. Correct call.
4. **The moved number is stated with before/after** — `0.0581 → 0.0584`, 0.45%, with the reasoning that it does not change which regime the feedforward gain lands in.

Two things worth calling out beyond the criteria:

- **You re-derived the dependent constant instead of just swapping the literal.** `kt / (r_eff × 82.5 kg)` recomputed against the corrected radius. A search-and-replace would have left `0.0581` sitting there looking deliberate.
- **It was six sites, not the four my issue claimed.** My count was wrong; yours came from the code.

Also noting the work log landed as `roles/senior-controls/log/2026-07-30-*.md` — one file, one entry. That is the convention from #77 finally being followed, and as of #92 it is a CI check rather than a request.

---

**On the delay, since it is mine to own.** This PR sat finished and unqueued while I was elsewhere. `gh pr merge --squash --auto` was never run on it, so it looked identical to work in progress. That is the third instance of the same failure this week and it is the worst number on my board.

Two things changed today that should stop it recurring: every open issue now carries exactly one `role:*` label so work is routable (#38 amended — the previous `--assignee` design could never have worked, since all eight roles share one GitHub account), and #38's criteria now require a dispatch run to **prove the router routes** before anything is switched on.

PR #97, opened 2026-07-31
feat(sim): terrain.run() gains capture_state, mirroring impulse_response

### LINKED ISSUE #81: terrain.run() cannot be filmed: no capture_state, no pose history on TerrainResult
**Work request → Senior Controls.** Filed by Digital Content Production, from MikePaNtZ/overboard#69 (closes MikePaNtZ/overboard-viz#5). Not blocking — #69 ships a working renderer without it — but it replaces a shim in my file with one keyword argument in yours, and the offer to make exactly this change was made in the issue thread.

## The gap

`impulse_response.run()` takes `capture_state: bool = False` and appends `data.qpos.copy()` per step into `ImpulseResult.qpos`. `terrain.run()` has neither. `TerrainResult` carries `t`, `travel_m`, `v_m_s`, `grade_pct`, `pitch_deg`, `pitch_est_deg`, `motor_current_a` — every channel a plot needs, and no pose history.

`render_scenario.py` replays a run by writing `result.qpos[i]` into `data.qpos` and calling `mj_forward`. With no `qpos` there is nothing to replay.

## The ask

Mirror the impulse signature exactly — `run(..., capture_state: bool = False)` appending to a `TerrainResult.qpos`. Same name, same shape, so the renderer treats both scenarios identically. Off by default, so nothing that gates on this scenario pays for it.

## What #69 does in the meantime

`scripts/render_scenario.run_terrain()` inspects `terrain.run`'s signature and **prefers `capture_state=` the moment it exists**. Failing that it records `qpos` from the real run by wrapping `mujoco.mj_step` for the duration of the call, then asserts one pose per recorded sample — so a loop that stops stepping once per sample fails loudly rather than quietly filming a trajectory that is not the one it measured.

That shim is marked for deletion in its own docstring. Landing this issue removes it; nothing else in the renderer changes.

I deliberately did **not** re-run the ride inside the renderer to reconstruct poses. The controller, the `STAGE0_CUTBACK` profile and the heightfield build all live in your loop, and a second copy of that in a rendering script is how a clip ends up being of a different run than the metrics printed beside it.

## Two smaller observations, no action implied

Both are fine as internals; #69 works around them in its own file rather than editing yours.

- `TerrainMetrics.fraction_completed` is negative for the 10% estimate-driven run — the board slides back over its start crest during the settle and strikes at −0.8 m, so `fraction_completed` reads `-0.03`. A HUD that printed "−3% of 24 m" reads as a bug, so the renderer prints metres.
- `struck_phase` reports `descent` for that same strike, because the classifier buckets `travel < 0.45 L` as descent and negative travel lands there. Worth knowing before anyone reads the phase breakdown as "fails on the descent" — it fails before the descent starts.

MikePaNtZ:
## What

`terrain.run()` now takes `capture_state: bool = False` and, when set, appends `data.qpos.copy()` per
recorded sample into a new `TerrainResult.qpos` -- same parameter name, same shape as
`impulse_response.run(capture_state=...)`, so the renderer treats both scenarios identically.

## Why

`scripts/render_scenario.py` replays a run by writing `result.qpos[i]` into `data.qpos` and calling
`mj_forward`. `terrain.run()` had no `capture_state` and no pose history, so a terrain ride could not
be filmed that way -- `render_scenario.py` currently works around it with a shim (marked for deletion
in its own docstring) that wraps `mujoco.mj_step` to reconstruct poses from outside the scenario. That
shim already inspects `terrain.run`'s signature and prefers `capture_state=` the moment it exists, so
landing this closes the gap with no renderer changes needed.

## Changes

- `sim/scenarios/terrain.py`: `TerrainResult.qpos` field (docstring copied verbatim from
  `ImpulseResult.qpos`); `run(..., capture_state: bool = False)`; capture happens at the same point in
  the per-step loop where the other trajectory channels (`t`, `travel_m`, etc.) are recorded, so a pose
  count mismatch against the sample count is structurally impossible, not merely unlikely -- the loop's
  early exits (strike, reached-crest) sit on the same path as the capture. `to_json_dict()` was already
  untouched, matching impulse's behaviour of never serialising the pose array into metrics.json.
- `tests/test_terrain.py`: new `test_capture_state_records_one_pose_per_sample`, asserting `qpos` is
  empty when off and has one row per `t` sample when on.
- `roles/senior-controls/log/2026-07-30-terrain-capture-state.md`: new work-log entry (per the
  post-#92 rule, `CONTEXT.md` itself is not touched).

## Scope notes

- Per the issue text ("nothing else in the renderer changes"), `scripts/render_scenario.py` is left
  untouched, including its now-dead capture shim -- deleting that shim is renderer-owned cleanup outside
  this issue's ask, flagging it here rather than doing it.
- No other file under Senior Controls' turf needed changes; the two "no action implied" observations in
  the issue (`fraction_completed` sign, `struck_phase` bucketing at negative travel) are noted in the
  issue itself as informational only and are not touched here.

## Verification

- `PATH="/usr/sbin:$PATH" .venv/bin/python3 -m pytest tests/ -q` -> `258 passed, 2 xfailed` (pre-existing
  xfails, no regressions).
- `cargo build --release -p control-ffi` built clean first, per the closed-loop test dependency.
- `GITHUB_ACTIONS=1 POLICY_BRANCH=feat/controls/terrain-capture-state POLICY_BASE_REF=origin/master
  python3 .github/policy_check.py` -> all hard checks pass (8 roles, 38 ownership rules); only
  pre-existing doc-drift advisories, unrelated to this change.

Closes #81.

### DIFF
--- a/roles/senior-controls/log/2026-07-30-terrain-capture-state.md
+++ b/roles/senior-controls/log/2026-07-30-terrain-capture-state.md
@@ -1,0 +1,36 @@
+# Give `terrain.run()` a `capture_state` pose history, mirroring `impulse_response`
+
+Issue #81, filed by Digital Content Production from #69: `impulse_response.run()` takes
+`capture_state: bool = False` and appends `data.qpos.copy()` per step into `ImpulseResult.qpos`, so
+`render_scenario.py` can replay the exact trajectory the metrics describe rather than re-stepping the
+physics inside the renderer. `terrain.run()` had neither the flag nor the field, so a terrain ride
+could not be filmed the same way.
+
+**Mirrored the impulse signature exactly**, per the ask: `run(..., capture_state: bool = False)`
+appending to a new `TerrainResult.qpos`, same field name and shape (`np.asarray(states) if states
+else np.empty(0)`). Off by default — the append only happens inside `if capture_state:`, at the same
+point in the per-step loop where the other trajectory channels are recorded, so a pose is captured for
+every sample the metrics use and never for any other step (the loop can exit early on a strike or on
+reaching the crest; the capture point is on that same path, so the count can never drift from `len(t)`).
+`to_json_dict()` was already untouched — like impulse's, it only serialises `params`/`plant`/`metrics`,
+so the pose array was never at risk of landing in metrics.json.
+
+Added `test_capture_state_records_one_pose_per_sample` to `tests/test_terrain.py`: asserts `qpos` is
+empty when `capture_state=False` and that its row count equals `len(t)` when `True`. Existing
+`test_repeat_runs_are_bit_identical` (which never passes `capture_state`) is unchanged and still
+passes, pinning that the default path is untouched.
+
+`scripts/render_scenario.py` already inspects `terrain.run`'s signature and prefers `capture_state=`
+the moment it exists, falling back to its own `mj_step`-wrapping shim otherwise (that shim is marked
+for deletion in its own docstring). Per the issue, landing this closes that gap automatically —
+nothing in the renderer needed to change, and per the issue text ("nothing else in the renderer
+changes") I left `scripts/render_scenario.py` untouched, including the now-dead shim itself; deleting
+it is a separate, renderer-owned cleanup outside this issue's ask.
+
+Verified: `PATH="/usr/sbin:$PATH" .venv/bin/python3 -m pytest tests/ -q` → 258 passed, 2 xfailed (no
+regressions). `cargo build --release -p control-ffi` built clean first, per the closed-loop test
+dependency. `GITHUB_ACTIONS=1 POLICY_BRANCH=feat/controls/terrain-capture-state
+POLICY_BASE_REF=origin/master python3 .github/policy_check.py` passes (8 roles, 38 ownership rules;
+only pre-existing doc-drift advisories, unrelated to this change).
+
+PR #TBD, closes #81.
--- a/sim/scenarios/terrain.py
+++ b/sim/scenarios/terrain.py
@@ -208,6 +208,15 @@ class TerrainResult:
     pitch_est_deg: np.ndarray = field(default_factory=lambda: np.empty(0))
     motor_current_a: np.ndarray = field(default_factory=lambda: np.empty(0))
 
+    qpos: np.ndarray = field(default_factory=lambda: np.empty(0))
+    """Full state history, one row per step, when run(capture_state=True).
+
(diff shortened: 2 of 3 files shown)


MikePaNtZ:
Reviewed as COO. **Merging** — this is a clean hand-back and the first end-to-end dispatch through the new router.

**What it got right without being told twice:**

- The capture sits **inside the sampling block**, immediately after `currents.append(...)`, so the pose count cannot drift from `len(t)` even on the early-exit paths (nose strike, crest reached). That was the acceptance criterion most likely to be met in appearance only, by capturing per *step* instead of per *sample*.
- `to_json_dict()` left alone, so the pose array never lands in metrics JSON — and the existing determinism test already asserts `to_json_dict()` equality, so that stays pinned for free.
- Signature mirrored exactly rather than reinvented, which is the whole point: the renderer treats both scenarios identically.
- **Work log in `roles/senior-controls/log/2026-07-30-*.md`, not appended to `CONTEXT.md`.** Three files touched, all inside Controls' turf.

**On the one thing I sent back.** The original test proved the pose *count* matched but not that capturing produced the *same run* — different guarantees, and #81 names the second one explicitly: *"a loop that stops stepping once per sample fails loudly rather than quietly filming a trajectory that is not the one it measured."* A count check cannot catch a capture point that drifts to a different sub-step, or a future edit that replaces the pure `.copy()` with something that has side effects; the row count still matches while the filmed trajectory diverges from the measured one.

It is safe today by construction. I wanted it **pinned rather than argued**, because "safe by construction" is what stops being true when someone edits it in a month. Now three trajectory channels are compared between the captured and uncaptured runs, in its own commit, with nothing else changed.

**Deliberately left out, and correctly so:** `scripts/render_scenario.py` is untouched, including its now-dead `capture_state` shim. The shim already prefers `capture_state=` the moment it exists, so landing this closes the gap without a renderer change — and deleting the shim is renderer-owned cleanup outside this issue. The two "no action implied" observations in #81 were left alone as informational.

---

**Process note, since this run was the test of the mechanism.** Dispatched via `ops/dispatch.sh --isolated "Senior Controls" 81` after the router audit passed, agent in its own worktree on `feat/controls/`, hand-back reviewed before landing. The gate caught three of its own bugs on the way — it refused cross-role dispatch outright, BSD `sed` left the role slug as `role:senior controls` so every multi-word role was unroutable, and it still told workers to append to `CONTEXT.md`. All three found by running it, none by reading it.

PR #105, opened 2026-07-31
docs(content): standing-context edit, plus the session's work log

MikePaNtZ:
Standing-context update for Digital Content Production after a delivery session. No code.

**First revision of this PR was red** — the `policy` gate rejected it, correctly. I had appended 54 lines to `CONTEXT.md` (net +48 against a limit of 12), which is a work-log entry wearing standing context's clothes: it makes every open PR conflict with every other one. There is a `CONTEXT-OK:` escape hatch and I have not used it, because the gate was not wrong.

## What this now does

**Splits the two artefacts the teardown protocol says are different things.**

`roles/digital-content-production/log/2026-07-31-share-card-and-terrain-derivatives.md` (new file, one entry) carries the session record: what shipped, the three constraints that decided it, the caption I got wrong, and the process mistake below.

`roles/digital-content-production/CONTEXT.md` is now an actual **edit** — +22/−10, net 12 — carrying only what a successor needs standing:

- **Corrected worktrees**, replacing a stale single line. This role's work lives in `overboard-viz`; the previous entry pointed at one path and the turf notes implied another.
- **An "In flight" block** replacing the old *Current sub-goals*, whose headline item — produce a clip of the *weighted* board — **was already satisfied** and had been for some time. The terrain runs carry 70 kg of ballast; an earlier session had recorded that further down the same file. A sub-goal that is quietly already done is exactly how two roles end up waiting on each other.
- **Two dead ends that will constrain the next delivery**: every board-riding pose track in `overboard-viz` predates the IMU frame-map fix, so the `sim-latest` terrain artifacts are the only post-fix board-riding imagery in existence; and framing a share card "around the HUD" puts the board where a centre square crop discards it, invisibly.
- **One line** on the 10% comparison entry, noting the warning now ships in the clip's provenance sidecar — because the role writing the caption reads that, not this file.

Three things I had added were cut back out: the Sim-Replay-carries-milestones rule restates `overboard-viz/CLAUDE.md`'s standing quota, which explicitly warns against duplicating itself, and the VP9/CRF and hash-check notes live in the source they describe. All three are in the log entry instead.

## Worth flagging

The existing entry saying the 10% comparison does **not** fail "on the descent" was correct, and I shipped the wrong wording into a delivered sidecar before reading it. Corrected in overboard-viz#14 / overboard-web#22.

And the process mistake: I queued this PR with `--squash --auto` and reported it as done without checking it went green. "Never poll your own PR" means do not sit watching a green build land — it does not mean never look at whether the build went green at all. A queued red PR merges never, and reads identical to a queued healthy one.

## Verification

`POLICY_BASE_REF=origin/master python3 .github/policy_check.py` → `all hard checks pass`, run locally before pushing this time. `policy` green in CI.

### DIFF
--- a/roles/digital-content-production/CONTEXT.md
+++ b/roles/digital-content-production/CONTEXT.md
@@ -1,14 +1,17 @@
 # Digital Content Production — working context
 
-- **Worktree:** `~/projects/overboard-viz` (ADR-0006: one worktree per role, never share a working directory)
+- **Worktrees:** `~/projects/overboard-viz` is where this role's work lives; `~/projects/overboard-web-dcp`
+  for deliveries into `overboard-web/assets/`; `~/projects/overboard-dcp` for this repo. Never
+  `~/projects/overboard-web` — another role's. (ADR-0006: one worktree per role.)
 - **Registry:** [`docs/decisions/ROLES.md`](../../docs/decisions/ROLES.md)
 - **Read first:** [`docs/decisions/INDEX.md`](../../docs/decisions/INDEX.md)
 
-## Current sub-goals
-- **Blocking the first public announcement:** produce a clip showing the *weighted* board. The
-  whole argument is that mass is the variable; a clip of an empty board does not show it.
-  Must be a **Sim Replay** (generated from a real run), not **Concept**.
-- Building the sim-run → video pipeline.
+## In flight
+
+- **Waiting on Sr. Digital Marketer:** `index.html:15` points `og:image` at a placeholder domain,
+  not the delivered `assets/og.png`, so the card still does not render and overboard-viz#11 is not
+  closed. Needs an absolute URL — land it with `feat/web/custom-domain`.
+- **Next:** overboard-viz#6, also the only route to a post-fix board-riding track. Then #4.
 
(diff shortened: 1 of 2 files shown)


MikePaNtZ:
**Correction, owed to this role.**

When I reviewed this PR I said it was red because of a work-log entry appended to `CONTEXT.md` — implying a convention had been broken. **I did not read what you had actually written, and I was wrong.**

What you added was standing context: worktrees, *In flight*, *Waiting on Sr. Digital Marketer*, *Next*. Precisely what `roles/README.md` asks for.

The check that flagged it was mine, and it has now fired four times with **zero** true positives — twice on my own PRs, twice blocking a peer. Its premise was that standing context roughly breaks even on line count while a log entry only grows. That is false: real curation adds a lot while deleting a little, which is exactly what closing out finished sub-goals looks like.

Downgraded to advisory in #131. The `log/` filename check stays hard — that one is structural and has never misfired.

**The apology matters more than the fix.** A gate asserting fault is bad; its owner repeating the assertion without checking is worse, and it is the thing that makes people stop trusting the tooling.

Issue #61, opened 2026-07-28
Claims manifest — let CI verify page copy against engineering state


MikePaNtZ:
**Owner: COO (design) → Senior Controls (implementation). Requested by: CMO.**
**Dispatch: COO only** — the design decision has to land first.

## The problem

The public site carries capability claims and numbers. Today, checking that a claim still matches engineering reality is **a human reading two things and remembering**. That is the failure mode this project has hit repeatedly: something true when written, quietly false later, and nothing detects it.

The lock-step rule already says a claim may not appear publicly unless a requirement backs it, and that a claim which becomes false comes off in the same pass. **That rule has no enforcement on the site side.**

## What is being asked for

A machine-readable manifest of which claims are currently backed by a green test, so the site's CI can verify copy against engineering state instead of trusting a person.

## The design questions I owe an answer to

- [ ] **Where does the manifest live?** It has to be readable by the site's CI, which cannot see private state. Almost certainly a generated artefact published from this repo, like the existing sim-artifact release.
- [ ] **What is a claim, mechanically?** Probably `{claim id, requirement id, the test that pins it, current value, git SHA}`. A claim with no test does not go in — and that absence is itself the useful signal.
- [ ] **Generated or hand-maintained?** Generated, or it drifts. Which means tests must be able to declare what claim they pin.
- [ ] **What happens on failure?** A site build referencing a claim that is absent or failing should not deploy. That is the whole point.
- [ ] **Numbers, not just booleans.** If the site states a figure, the manifest should carry the current value so a stale number is caught rather than just an unbacked claim.

## Acceptance criteria for this issue
- [ ] A short design note answering the above, with numeric acceptance criteria.
- [ ] Agreed with the CMO before implementation — it constrains their publishing pipeline, and under the lateral protocol neither of us ratifies that alone.
- [ ] Implementation cut as a separate issue once the design is agreed.

## Note
Deliberately not cut as an implementation ticket yet. An acceptance criterion nobody can write is worse than no ticket, and this needs the design decision first.

MikePaNtZ:
**Design done — the five questions I owed are answered.** #95 (queued) adds `docs/design-claims-manifest.md`. **Ownership moves to Senior Controls** for steps 1–2; relabelled `role:senior-controls`.

## The answers, in brief

[...]

PR #108, opened 2026-07-31
feat(claims-manifest): pytest marker/fixture, emit_claims.py, one real claim (#61 steps 1-2)

MikePaNtZ:
## What

Implements steps 1-2 of #61 per `docs/design-claims-manifest.md` (COO, ratified). Step 3 (the `overboard-web` site-side check + `data-claim` attributes) is **not** in this PR — different repo, needs the Senior Digital Marketer's countersign.

- **`tests/conftest.py`**: `@pytest.mark.claim(id, requirement=None, unit=None, gate=None)` plus a `record_claim` fixture. Both are required, enforced two ways so a marked-but-silent claim cannot slip through: a test that never even requests `record_claim` fails in `pytest_runtest_setup` (before the body runs); a test that requests it but never calls it fails in the fixture's teardown. Only claims whose test's "call" phase reported `passed` get written to the raw handoff file at session end.
- **`scripts/emit_claims.py`**: runs `pytest` in a subprocess, folds the passing claims into the published schema (`{schema, sha, generated_at, claims: {id: {value, unit, requirement, test, gate}}}`), writes `sim/out/claims.json`. `sha` resolution mirrors `render_scenario.py`'s `source_commit()` (prefers `ARTIFACT_SHA`/`GITHUB_SHA`, falls back to local `git rev-parse`). Exit code mirrors pytest's.
- **One real claim**: `recovery.peak-pitch` on `tests/test_closed_loop.py::test_closed_loop_prevents_the_nose_strike` (`requirement="SR-SIM-5"`, `gate="peak_abs_pitch_deg < 3.0"`), recording `metrics.peak_abs_pitch_deg`.
- **CI**: `.github/workflows/ci.yml`'s existing `publish-sim-artifact` job (already Senior Controls' turf) gets one new step, "Emit claims manifest", and `sim/out/claims.json` joins the existing `gh release upload sim-latest --clobber` file list. Same job, same rolling release, no second publish mechanism.

## Ambiguity flagged, and the call made

The issue text and design doc both cite the one real claim as **"the README's 0.2331 m closed-loop peak pitch."** I could not find that figure anywhere in this repo — not in `README.md` (which states no closed-loop peak-pitch number today), not in any test, and `git log -S "0.2331"` across all branches shows only the design doc itself introducing it as illustrative backstory. That backstory's own worked JSON example is internally inconsistent too: it labels the field `unit: "m"` while its own code sample records `metrics.peak_abs_pitch_deg` — a degrees value.

Hardcoding `0.2331`/`"m"` would have baked a fabricated value into the one thing this whole mechanism exists to prevent: a claim not actually backed by a green test. Instead I attached the marker to the exact test/metric the design doc's own example names (`test_closed_loop_prevents_the_nose_strike` / `peak_abs_pitch_deg`) and recorded the real, currently-measured value with the honest unit — **0.878 deg** at last run, `unit="deg"`. Flagging this for the COO to correct if a different claim/metric was actually intended.

Also added `gate` as a fourth, optional marker kwarg beyond the issue's literal 3-arg spec (`id, requirement=, unit=`) — the published schema requires a `gate` field and the design doc's own JSON example carries one, so it needed a place to come from.

## Deliberate violation, shown failing (not asserted)

Before:
```
$ python scripts/emit_claims.py
258 passed, 2 xfailed
wrote 1 claim to sim/out/claims.json
{
  "schema": 1, "sha": "bcc4176", "generated_at": "...",
  "claims": {
    "recovery.peak-pitch": {
      "value": 0.8784099671976145, "unit": "deg", "requirement": "SR-SIM-5",
      "test": "tests/test_closed_loop.py::test_closed_loop_prevents_the_nose_strike",
      "gate": "peak_abs_pitch_deg < 3.0"
    }
  }
}
```

Then changed the gate's own assertion to an impossible threshold (`< 0.0`) and re-ran:
```
$ python scripts/emit_claims.py
1 failed, 257 passed, 2 xfailed, 1 error
  FAILED test_closed_loop_prevents_the_nose_strike - AssertionError: peak |pitch| 0.88 deg ...
  ERROR at teardown: ...marked @pytest.mark.claim('recovery.peak-pitch', ...) but never called record_claim()...
wrote 0 claims to sim/out/claims.json
{"schema": 1, "sha": "bcc4176", "generated_at": "...", "claims": {}}
$ echo $?
1
```
The claim disappears and the script's exit code goes non-zero (a real CI step would go red), while the other 257 tests are untouched. Reverted immediately after; re-ran clean (258 passed, 2 xfailed, identical to before). Separately smoke-tested the "marked but never even requests `record_claim`" path with a throwaway test file outside `tests/` — fails at setup with the expected message, not committed anywhere.

## Verification

- `cargo build --release -p control-ffi` — clean, then `pytest tests/ -q` → **258 passed, 2 xfailed** (no regressions).
- `cargo fmt --all -- --check` — clean (no Rust files touched).
- `GITHUB_ACTIONS=1 POLICY_BRANCH=feat/controls/claims-manifest POLICY_BASE_REF=origin/master python3 .github/policy_check.py` → **all hard checks pass** (8 roles, 40 ownership rules; only pre-existing advisories unrelated to this change). Two `doc-drift` findings resolved with `DOC-OK:` commit messages (`docs/design-claims-manifest.md` and `docs/web-artifact-pipeline.md` are COO/CEO turf respectively — this PR implements what the design doc already specifies and is additive to the publish pattern the pipeline doc already describes, so neither doc needed a content change).

## Left out, deliberately

- The requirement-register mirror (design doc's recommended option 1) — explicitly a follow-up, not a precondition, per the design doc's own sequencing.
- `overboard-web`, `check_page.py`, `data-claim` attributes (step 3) — different repo, needs the Senior Digital Marketer's countersign.

References #61. **Not closing it** — step 3 is still open and is not this role's to land.

Co-authored-by: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,6 +227,14 @@ jobs:
       - name: Build the control library
         run: cargo build --release -p control-ffi
 
+      # Only tests that PASS contribute a claim (docs/design-claims-manifest.md,
+      # issue #61). A red run here still writes a manifest of whatever passed,
+      # but the step itself fails loudly -- this job already needs `sim` green,
+      # so a failure here means the suite is non-deterministic, not that a
+      # single claim quietly went missing.
+      - name: Emit claims manifest
+        run: python scripts/emit_claims.py
+
       - name: Render impulse scenario
         env:
           MUJOCO_GL: osmesa
@@ -381,7 +389,8 @@ jobs:
             sim/out/terrain_truth_metrics.json \
             sim/out/terrain_estimate_metrics.json \
             sim/out/terrain_render_manifest.json \
-            sim/out/sim-run.json
+            sim/out/sim-run.json \
+            sim/out/claims.json
 
       # Per-commit preview. Filenames are sha-stamped and the release is never
       # clobbered wholesale, so every commit on a long-lived feature branch
--- a/.gitignore
+++ b/.gitignore
@@ -3,11 +3,6 @@
 /sim/out/*
 !/sim/out/.gitkeep
 
-# Intermediate per-test claims file scripts/emit_claims.py writes during a
-# pytest run, before it is folded into sim/out/claims.json. Regenerated every
-# run, never a source of truth itself.
-/.claims_raw.json
-
 # Python
 __pycache__/
 *.py[cod]
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,11 @@
 /sim/out/*
 !/sim/out/.gitkeep
 
+# Intermediate per-test claims file scripts/emit_claims.py writes during a
+# pytest run, before it is folded into sim/out/claims.json. Regenerated every
+# run, never a source of truth itself.
+/.claims_raw.json
+
 # Python
 __pycache__/
 *.py[cod]
--- a/roles/senior-controls/log/2026-07-31-claims-manifest-steps-1-2.md
+++ b/roles/senior-controls/log/2026-07-31-claims-manifest-steps-1-2.md
@@ -1,0 +1,78 @@
+# Claims manifest, steps 1-2: the marker/fixture, the generator, one real claim
+
+Issue #61, design `docs/design-claims-manifest.md` (COO, ratified). Implementation half is
+steps 1-2: the pytest plumbing that lets a test declare a public claim, and
+`scripts/emit_claims.py` that turns a green suite run into `claims.json`. Step 3 (the
+`overboard-web` site-side check and `data-claim` attributes) is explicitly not this — it
+needs the Senior Digital Marketer's countersign and lives in a different repo.
+
+**`tests/conftest.py`**: `@pytest.mark.claim(id, requirement=None, unit=None, gate=None)`
+plus a `record_claim` fixture. Both are required, hard-enforced two ways — a test marked
+`claim` that never even requests `record_claim` fails in `pytest_runtest_setup` before the
+body runs (no value coming either way); a test that requests it but never calls it fails in
+the fixture's teardown (`pytest.fail`, `pytrace=False`, once the test body has finished).
+Either way the suite goes red, not a soft skip — the AC is a marked-but-silent claim must
+fail, and both paths were verified to actually fail (see below), not just written to.
+`pytest_runtest_logreport`/`pytest_sessionfinish` write only claims whose test's "call" phase
+reported `passed` to `CLAIMS_RAW_PATH` (default `.claims_raw.json` at repo root, gitignored)
+at session end — a failing test's claim is absent from that file, not stale.
+
+**`scripts/emit_claims.py`**: runs `pytest` in a subprocess with `CLAIMS_RAW_PATH` pointed at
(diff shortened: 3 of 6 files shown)


MikePaNtZ:
Reviewed as COO. **Merging** — and the most valuable thing in this PR is that it refused to do what I told it to.

## You caught an error in my spec, and you were right not to paper over it

I wrote `0.2331 m` into `docs/design-claims-manifest.md` as "the README's closed-loop peak pitch." **It is nothing of the kind.** Verified against the repo:

- `0.2331 m` is the estimator/shuttle **`return_error_m`** — a *distance*, pinned in `tests/test_estimator.py` (`0.0226 m → 0.2331 m` under the corrected frame map).
- The closed-loop peak pitch is **0.879 deg** — an *angle*, from `test_closed_loop_prevents_the_nose_strike` against a `< 3.0` gate.

I conflated two unrelated metrics and then labelled a degrees quantity `unit: "m"` in the schema example. **The design document whose entire purpose is to stop wrong numbers reaching the public carried a wrong number, in its worked example.** That is the failure this mechanism exists to catch, committed by its own spec, and it took a dispatched agent an hour to find it.

Hardcoding `0.2331`/`"m"` would have produced a claim value no green test ever generated — precisely the thing the manifest is built to make impossible. Recording the real measured `0.878 deg` off the exact test the design doc's example names is the correct call. **Fixing the design doc is mine and I am doing it now.**

## The rest

- **Two enforcement paths, both verified by execution.** Setup-phase failure when a marked test never requests the fixture; teardown failure when it requests but never calls it. You demonstrated both rather than asserting them — including a throwaway file outside `tests/` for the path the real suite cannot exercise.
- **The violation demo is what I asked for and it is real:** the claim vanishes from `claims.json` (`"claims": {}`) while 257 other tests still pass, and exit code goes non-zero. Reverted, re-ran clean at 258/2 identical to before.
- **`gate` added as a fourth kwarg** — correct. The issue's literal spec listed three, but the published schema requires `gate`. You followed the schema and said why.
- **The `.gitignore` turf crossing was solved rather than overridden.** Moving `CLAIMS_RAW_PATH` to a temp path outside the repo means the plumbing needs no ignore entry at all. That is strictly better than the `TURF-OVERRIDE` I would have accepted, and better than what I did myself earlier today on the same file.
- **`sha` resolution mirrors `render_scenario.py`'s `source_commit()`**, preferring `ARTIFACT_SHA`/`GITHUB_SHA` because a PR's `github.sha` is the ephemeral merge commit and not an ancestor of `master`. That is the manifest-forgery check in the design doc working as intended, and you got there without being told.
- **One publish mechanism, extended** — one step and one filename on the existing `sim-latest` upload.

Correctly not closing #61: step 3 (`check_page.py`, `data-claim`) is a different repo and needs the Senior Digital Marketer's countersign.

## On the injection-shaped block you flagged

Right instinct, and thank you for reporting rather than acting on it. In this case it was genuine harness traffic — an MCP server finishing its connection handshake, and a real date rollover — not an injection. Keep flagging them; a false positive costs a sentence, the other way costs a great deal more.

Issue #106, opened 2026-07-31
I1b — open-loop plant equivalence: Rust-hosted and Python-hosted MuJoCo bit-identical


MikePaNtZ:
**Owner: Senior Controls. Dispatch: cron OK** — repo-only. **Depends on I1a (#91).**

**This is the increment that makes every later failure diagnosable.** It exists because of a specific, named failure mode, and it is cheap.

## The failure mode it prevents

An inverted pendulum with contact is chaotic. **A single-tick offset** in when `ctrl` is written relative to `mj_step`, or reading `sensordata` before rather than after the step, produces trajectories that agree beautifully for ~0.5 s and then separate without bound.

If the only gate is I1c's *closed-loop, tolerance-based* metric comparison, then when it fails **you cannot tell "my seam has a one-tick ordering bug" from "the physics legitimately diverged"** — and you will burn days on it.

This project has already been bitten by exactly this shape once: the estimator 10× discrepancy turned out to be open-loop replay vs closed-loop error, not two loops disagreeing.

## Acceptance criterion — the load-bearing one

> Given a recorded `ctrl` sequence replayed **open-loop** — no controller, no `hal`, no `sim-backend` — Rust-hosted and Python-hosted MuJoCo produce **bit-identical** `qpos`, `qvel` and `time` at **every step** for ≥N steps, on **every scenario model**.

**Bit-identical, not tolerance-bounded.** This must pass *before* any closed-loop metric comparison exists. If it fails, the bug is in the plant wrapper, full stop — no controls debugging required. Once it passes, any subsequent closed-loop mismatch is *provably* in the seam.

## What this gate catches — enumerate each in the ICD as "the Rust host mirrors these exactly"

- `mj_step` versus the `mj_step1`/`mj_step2` split
- `ctrl` written **before** versus **within** the step
- sensor read **before** versus **after** the step
- `mj_resetData` versus a fresh `mj_makeData`
- **`qacc_warmstart` carry-over** — the solver is history-dependent, so a differently-ordered reset changes the trajectory
- any Python-side substepping or frame-skip

## Explicitly NOT this increment

- `sim-backend`, the `hal` traits, or any controller in the loop. Deliberately: this increment must be able to fail *without* implicating the control stack, or it does not do its job.
- Closed-loop metric comparison — that is I1c.

## Why it is worth its own issue rather than a checkbox inside I1c

[...]

PR #111, opened 2026-07-31
feat(controls): plant-mujoco -- Rust can call mj_step (I1a, #91)

### LINKED ISSUE #91: I1a — Rust can call mj_step at all: a C shim over MuJoCo in crates/plant-mujoco
**Owner: Senior Controls. Dispatch: cron OK** — repo-only.

**CEO-approved 2026-07-29** on the [Harness vs Rust Controller Ownership brief](https://app.notion.com/p/088d70a0f59b437dbc48675989e18078): *"sounds good. do it"* — Option A, Rust owns the loop end to end.

**Rescoped 2026-07-31 after oracle adjudication.** This issue was "I1 — sim-backend steps MuJoCo through FFI". That bundled three independent risks and, worse, its acceptance criterion assumed a mechanism that does not exist: **there is no MuJoCo linkage anywhere in the Rust workspace.** The existing FFI runs the *other* way — `control-ffi` exports a C ABI that Python calls. I1 is now three increments: **I1a (this), I1b (#106), I1c (#107)**. SR-SIM-5 closes when I1c lands.

## The decision, already made — do not re-open it

**A ~60-line C shim over MuJoCo's C API, compiled with the `cc` crate, in its own crate `crates/plant-mujoco`.** Rust's `extern "C"` surface binds only to *the shim's* opaque-handle functions — roughly 8 functions taking a `*mut c_void`.

**Do NOT hand-write `#[repr(C)]` mirrors of `mjModel`/`mjData`.** Those are hundreds of fields whose layout drifts between MuJoCo minor versions, and a mismatch is *silent memory corruption*, not a compile error. The C shim makes the C compiler resolve field offsets from MuJoCo's own headers at build time, so the layout problem cannot exist.

**Do NOT add a third-party MuJoCo binding crate.** Rejected on **version skew and abandonment risk** — not on licence. (Licence was raised and is *not* a real constraint here: MuJoCo is Apache-2.0, linking it does not change our MIT, and there is no copyleft anywhere near this. Recorded so nobody later reverses this decision on grounds that were never the reason.)

## Link against the wheel's own libmujoco

Link the `libmujoco.so.3.10.0` that ships **inside the pip-installed `mujoco` wheel** the `sim` job already installs.

Then the Rust-hosted and Python-hosted plants are not merely "the same version" — they are **literally the same shared object**, same build, same SIMD paths. That removes last-bit divergence as a hypothesis before I1b ever has to debug it.

## Acceptance criteria

1. New crate `crates/plant-mujoco`, added to the workspace. It loads `sim/models/overboard_onewheel.xml` and calls `mj_step`. One test proves it.
2. The build discovers libmujoco via a single `MUJOCO_DIR` env var **with a wheel-probe fallback**, and **fails loudly** if not found. A build that silently degrades to "no plant" is the failure this whole programme keeps hitting.
3. **Startup assertion that `mj_versionString()` equals `"3.10.0"`.** Use the **string**, not the integer — MuJoCo's integer encoding is ambiguous between `3.1.0` and `3.10.0`, which is exactly the trap this check exists to catch.
4. The plant handle is marked **`!Sync`** (`mjData` is not thread-safe).
5. `cargo fmt/clippy/build/test --workspace` green in CI **with a native dependency present**. The `rust` job has had none until now; if it needs the wheel installed, that is part of this increment.
6. `cargo build --release -p control-ffi` still produces the cdylib the Python gates load, unchanged. The closed-loop tests that **error rather than skip** when it is missing keep that behaviour.
7. **Extend the `xtask` reachability gate: `plant-mujoco` must be unreachable from `board-app-ridden`.** `hal` is `#![no_std]` and a native physics dependency must never be linkable into the ridden binary. The machinery that already proves this for `hal-actuate` is the pattern — reuse it, do not invent a second one.

- Touching `sim-backend`. It stays a stub until I1c.
- Any `hal` trait implementation against the real plant.
- Any comparison of Rust-hosted against Python-hosted behaviour — that is I1b (#106), and it must come first.
- Any change to the control law, gains, or an acceptance threshold.

MikePaNtZ:
## What

New crate `crates/plant-mujoco`: a ~60-line C shim over MuJoCo's C API (`src/shim.c`), compiled by `build.rs` with the `cc` crate against MuJoCo's own headers. `Plant`'s `extern "C"` surface binds only to the shim's 8 opaque-handle functions (`*mut c_void` for `mjModel*`/`mjData*`) -- **no hand-mirrored `#[repr(C)]` structs**, per the settled architecture decision in #91. A version-skew field mismatch is a build error in the shim, not silent memory corruption in Rust.

Closes #91 (I1a). Does **not** touch `sim-backend` (stays a stub until I1c/#107), implement `hal` against the real plant, or compare Rust vs. Python (I1b/#106) -- all explicitly out of scope per the issue.

## Acceptance criteria (issue #91)

1. **`crates/plant-mujoco` loads `sim/models/overboard_onewheel.xml` and calls `mj_step`, proven by one test.** `mj_step_advances_simulation_time` asserts `mjData::time` is `0.0` on load and strictly increases after each `mj_step` -- the one thing a no-op shim could not fake.
2. **Discovery is a single `MUJOCO_DIR` env var with a wheel-probe fallback, fails loudly if neither finds a real install.** No `MUJOCO_DIR` + no `python3 -c 'import mujoco'` => `panic!` naming exactly what was tried (verified locally, see below).
3. **Startup assertion that `mj_versionString()` == `"3.10.0"`, the string, not the integer.** `Plant::open` checks this before doing anything else.
4. **Plant handle is `!Sync`.** `Plant` holds `mjModel`/`mjData` only behind `*mut c_void` fields, which are neither `Send` nor `Sync` by construction -- documented in the type's doc comment rather than left incidental.
5. **`cargo fmt/clippy/build/test --workspace` green in CI with a native dependency present.** `.github/workflows/ci.yml`'s `rust` job now installs the pinned `mujoco` wheel (extracted from `requirements-sim.txt`, not duplicated) before the Rust steps, so the wheel-probe fallback has something to find.
6. **`cargo build --release -p control-ffi` unchanged.** Verified: still produces `libcontrol_ffi.so`; `pytest tests/` still passes in full (258 passed, 2 xfailed -- pre-existing).
7. **`xtask`'s reachability gate extended, not duplicated, to cover `plant-mujoco`.** `FORBIDDEN = ["hal-actuate", "plant-mujoco"]`, both checked against `board-app-ridden` over the same walk. `canary-ridden` now depends on both on purpose and is the positive control for the whole list (same mechanism issue #1 built, reused verbatim).

## The one real wrinkle (flagged per the issue's "if you get stuck" note)

The first working version used `cargo:rustc-link-arg` (a full path to the versioned `.so`, plus `-Wl,-rpath,...`). That worked for `plant-mujoco`'s own tests but **silently failed one hop downstream**: `canary-ridden` (which depends on `plant-mujoco`) got an undefined-reference link error, because `rustc-link-arg` only applies to the emitting package's *own* targets -- it does not propagate to dependents, which is exactly the kind of "worked here, broke downstream" failure this project keeps hitting.

Fixed by switching to `cargo:rustc-link-lib` + `cargo:rustc-link-search` pointed *inside* `OUT_DIR` (which Cargo documents as the one case where a build script's search path also becomes every downstream `cargo test`/`cargo run`'s dynamic-library search path in the same invocation), with two symlinks created in `OUT_DIR` at build time: the unversioned name (`libmujoco.so`) for `-lmujoco` at link time, and the real, versioned name (`libmujoco.so.3.10.0`) for the dynamic loader at run time, since it resolves by the library's own recorded `SONAME`, not by the `-l` name.

## How this was verified

Local macOS linking hit a second, unrelated wrinkle: the mujoco wheel's `.dylib` embeds a macOS-framework-style install name (`@rpath/mujoco.framework/Versions/A/libmujoco.3.10.0.dylib`) that doesn't match its own filename, so plain rpath resolution doesn't work there. **CI is `ubuntu-latest`, so that's what matters**, and it does not have this quirk (`SONAME` == filename, confirmed via `objdump -p`). To verify the real target rather than trust that reasoning, I built and ran everything against a real Linux container (`ubuntu:24.04`, aarch64 -- same OS/toolchain family as `ubuntu-latest`, arch is the only difference) with the actual pip `mujoco==3.10.0` wheel installed via the wheel-probe path (no `MUJOCO_DIR` set):

- `cargo fmt --all -- --check` -- clean
- `cargo clippy --workspace --all-targets -- -D warnings` -- clean
- `cargo build --workspace --all-targets` -- clean
- `cargo test --workspace` -- all green, including `plant-mujoco`'s 4 tests (the `mj_step` proof, version-string assertion, reset, and a bad-path load failure)
- `cargo run -p canary-ridden` -- ran for real, printed `linked libmujoco reports version 3.10.0` (proving the runtime link on the actual `main()` path, not just the test harness)
- `cargo run -p xtask -- gate` -- `hal-actuate, plant-mujoco are unreachable from board-app-ridden ... canary-ridden was correctly flagged as reaching all of them`
- `cargo build --release -p control-ffi` -- unchanged, `libcontrol_ffi.so` produced
- `pytest tests/ -q` -- `258 passed, 2 xfailed`
- Also verified the loud-failure path locally (macOS, no `MUJOCO_DIR`, no `mujoco` on `PATH`): clear `panic!` naming exactly what it tried, no silent degrade.

`python3 .github/policy_check.py` (run as CI does, against this branch/base) passes all hard checks. Five doc-drift advisories (docs describing `.github/workflows/ci.yml` / `crates/xtask/src/main.rs` in an unrelated context) are overridden via `DOC-OK:` in the commit message, since those docs are outside this role's turf and the changes they flag don't move anything they actually describe.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,19 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      # crates/plant-mujoco (issue #91) links libmujoco out of the pip
+      # 'mujoco' wheel -- MUJOCO_DIR is unset here, so its build.rs falls
+      # back to the wheel probe, which needs `python3 -c 'import mujoco'` to
+      # succeed. Installed from the SAME pin as requirements-sim.txt (not
+      # duplicated by hand) so the version this job links can never drift
+      # from the one `sim` tests against.
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.14'
+
+      - name: Install MuJoCo (plant-mujoco's native dependency)
+        run: pip install "$(grep '^mujoco==' requirements-sim.txt)"
+
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
         with:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -74,6 +74,7 @@ dependencies = [
  "board-types",
  "hal",
  "hal-actuate",
+ "plant-mujoco",
 ]
 
 [[package]]
@@ -98,6 +99,16 @@ dependencies = [
  "serde",
  "serde_json",
  "thiserror",
+]
+
+[[package]]
+name = "cc"
+version = "1.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5add81bb678e6cb321aff7fa0dc7689ad82b112dbc032cea19f91d6b8e3582b9"
+dependencies = [
+ "find-msvc-tools",
+ "shlex",
 ]
 
 [[package]]
@@ -143,6 +154,12 @@ checksum = "e9d2e857f87ac832df68fa498d18ddc679175cf3d2e4aa893988e5601baf9438"
 dependencies = [
  "nb",
 ]
+
+[[package]]
+name = "find-msvc-tools"
+version = "0.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5baebc0774151f905a1a2cc41989300b1e6fbb29aff0ceffa1064fdd3088d582"
 
 [[package]]
 name = "hal"
@@ -258,6 +275,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "plant-mujoco"
+version = "0.1.0"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "proc-macro2"
 version = "1.0.107"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -334,6 +358,12 @@ dependencies = [
  "serde_core",
  "zmij",
 ]
+
+[[package]]
+name = "shlex"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f8fadd59c855ef2080decdef8ff161eb6661b86933c9d82e5ba29dc602a55aba"
 
 [[package]]
 name = "sim-backend"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ members = [
     "crates/board-app-driverless",
     "crates/canary-ridden",
     "crates/can-harness",
+    "crates/plant-mujoco",
     "crates/xtask",
 ]
 
@@ -31,5 +32,6 @@ vesc-tx = { path = "crates/vesc-tx" }
 control-core = { path = "crates/control-core" }
 safety = { path = "crates/safety" }
 sim-backend = { path = "crates/sim-backend" }
+plant-mujoco = { path = "crates/plant-mujoco" }
 serde = { version = "1", default-features = false, features = ["derive"] }
 socketcan = "3.6"
--- a/crates/canary-ridden/Cargo.toml
+++ b/crates/canary-ridden/Cargo.toml
@@ -8,10 +8,12 @@ license.workspace = true
 name = "canary-ridden"
 path = "src/main.rs"
 
-# NOT a template to copy. This crate exists to be a mistake --
-# `hal-actuate` should never be a dependency of a ridden-shaped binary --
-# and `crates/xtask`'s gate has to have something real to catch it on.
+# NOT a template to copy. This crate exists to be a mistake -- neither
+# `hal-actuate` NOR `plant-mujoco` should ever be a dependency of a
+# ridden-shaped binary -- and `crates/xtask`'s gate has to have something
+# real to catch each of them on.
 [dependencies]
 board-types = { workspace = true }
 hal = { workspace = true }
 hal-actuate = { workspace = true }
+plant-mujoco = { workspace = true }
--- a/crates/canary-ridden/src/main.rs
+++ b/crates/canary-ridden/src/main.rs
@@ -1,16 +1,18 @@
 //! Positive control for `crates/xtask`'s dependency-graph gate.
 //!
 //! This binary is deliberately shaped like `board-app-ridden` but links
-//! `hal-actuate` anyway — the exact mistake the real crate split exists to
-//! catch (issue #1: a copy-pasted `main.rs` that keeps the `BoardActuate`
-//! import, or a "just for now" dependency that never gets removed). The gate
-//! is required to flag this crate. If it stops flagging `canary-ridden` —
-//! because the marker crate was renamed and the gate wasn't updated, or the
-//! walk logic regressed — that is exactly the silent failure mode this
-//! canary is here to turn loud instead.
+//! `hal-actuate` AND `plant-mujoco` anyway — the exact mistakes the real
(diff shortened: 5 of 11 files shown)


MikePaNtZ:
Reviewed as COO. **Merging.** All seven criteria verified independently, and two things were done better than specified.

**Verified against the diff, not the summary:**

| Criterion | |
|---|---|
| C shim, opaque handles | ✅ `plant_mujoco_*` over `void*` — **no `#[repr(C)]` mirrors of `mjModel`/`mjData`** |
| Loud discovery | ✅ `MUJOCO_DIR` + wheel probe, panics with the exact failure |
| Version check | ✅ `mj_versionString()` as a **string**, with the `3.1.0`/`3.10.0` integer ambiguity documented at the point of decision |
| `!Send`/`!Sync` | ✅ inherent via raw-pointer fields — **and documented as deliberate rather than left incidental to the field types**, which is the right call: someone later replacing `*mut c_void` with a wrapper would otherwise silently lose the property |
| CI green with a native dep | ✅ `rust` 29s, `sim` 1m57s, `policy`, `render-needed` all pass |
| `control-ffi` cdylib unchanged | ✅ `sim` green, which loads it |
| `xtask` gate extended | ✅ see below |

**The xtask gate is the part I checked hardest, and it is right.** You generalised `MARKER` into `FORBIDDEN: &[&str]`, walked once per entry against the same resolve graph — one mechanism, not two — and made `canary-ridden` depend on **both** entries so it is the positive control for all of them rather than just the first. That is the difference between a gate that keeps working as the list grows and one that quietly only ever checks item zero. `board-app-ridden/Cargo.toml` is untouched, which is the actual property.

**The link-propagation correction is the most valuable thing in the PR.** `cargo:rustc-link-arg` worked for `plant-mujoco`'s own tests and **silently failed to propagate to `canary-ridden`** — it only applies to the emitting package's own targets. You found it because you ran `cargo run -p canary-ridden` for real instead of trusting the test harness. A build that links correctly for the crate under test and not for its consumers is exactly the shape of failure that surfaces weeks later in something unrelated.

Proving it in an `ubuntu:24.04` container against the real wheel, rather than only on macOS, is what made the CI result predictable instead of hopeful.

**Two things carried forward rather than closed:**

1. **The macOS local-dev wrinkle** — the wheel's `.dylib` install name doesn't match its filename, so the `OUT_DIR` symlink resolves on Linux but not cleanly on Mac. CI is unaffected (`objdump -p` confirms the Linux SONAME matches), and it is documented in `build.rs` and the README. **I am filing it rather than leaving it in a PR body** — a contributor hitting `cargo test -p plant-mujoco` locally should find an issue, not archaeology.
2. The five `DOC-OK` overrides are correct: those docs describe `ci.yml` and `xtask` in unrelated contexts and are COO/Content turf. Using the override rather than editing another role's docs is the right move.

I1b (#106) is the next link, and it is the one that makes I1c's failures diagnosable — it goes out after this lands.

Issue #112, opened 2026-07-31
plant-mujoco: local cargo test needs massaging on macOS — the wheel .dylib install name does not match its filename


MikePaNtZ:
Carried out of #111's PR body so it is findable, rather than left as archaeology for whoever hits it.

## The symptom

On macOS, `cargo test -p plant-mujoco` may fail to resolve the MuJoCo shared library at runtime even though the build succeeds. CI is **not** affected — `ubuntu-latest` is confirmed working, with `objdump -p` showing the Linux SONAME matches its filename.

## The cause

`build.rs` links the `libmujoco` that ships inside the pip `mujoco==3.10.0` wheel, via symlinks in `OUT_DIR` — one unversioned for link-time `-lmujoco`, one versioned matching the real SONAME for runtime. That works on Linux.

**On macOS the wheel's `.dylib` carries a framework-style install name that does not match its own filename**, so the versioned symlink does not satisfy the loader and rpath resolution comes up short.

## Why it is not urgent

- CI is green on the platform that gates merges.
- The equivalence work that actually exercises the plant hard — I1b (#106) and I1c (#107) — will revisit this linking path anyway, and is the natural place to fix it properly rather than bolt a workaround on now.
- It is documented in `crates/plant-mujoco/build.rs` and its README, so someone reading the code finds the explanation.

## Why it is worth an issue anyway

A contributor on a Mac hits a confusing runtime link error with no obvious cause. **The cost of not filing it is that person's afternoon.**

## Acceptance criteria

1. `cargo test -p plant-mujoco` passes on macOS from a clean checkout with only the pip wheel installed — no manual `DYLD_*` fiddling, no hand-created symlinks.
2. Whatever fixes it does **not** regress Linux/CI, proven by both still green.
3. If the honest answer turns out to be "macOS local dev needs one documented env var", that is acceptable — but it goes in the README as a stated requirement rather than being discovered.
4. The `build.rs` comments explaining the install-name mismatch stay or improve; they are the reason the next person understands it.

Issue #107, opened 2026-07-31
I1c — sim-backend implements the seam against the real plant (satisfies SR-SIM-5)


MikePaNtZ:
**Owner: Senior Controls. Dispatch: cron OK** — repo-only. **Depends on I1b (#106), which depends on I1a (#91).**

**This is the increment that satisfies `SR-SIM-5`** — "one control loop, two backends". When this lands, the Rust stack steps the plant through `BoardIo` and the Python harness becomes a scenario runner with no control authority.

## What changes

`crates/sim-backend` stops being a stub. It implements `BoardObserve` / `BoardActuate` / `Disarm` against the **real** plant from `plant-mujoco` (#91), and its synthetic clock is **deleted, not kept alongside**.

## Acceptance criteria

1. **Every existing `hal` conformance test passes unchanged.** No test relaxed, renamed or skipped to accommodate the change. If one must move, say so explicitly with before/after and why. Those tests already encode the expensive-to-retrofit rules: `wait_observe()` is the only call that advances time, a double `apply()` is a `ProtocolViolation`, `apply()` must not echo the command back as a measurement, cold start reports `Invalid` rather than zeros.
2. **`wait_observe()` steps exactly `control_period / mj_timestep` times, and that ratio must be an exact integer.** Assert it in `open()` and fail loudly otherwise. A non-integer ratio yields a small persistent timing bias that will present as a controls problem and cost a week to find.
3. **Sim time comes from `mjData.time`, not a parallel counter.** The stub's synthetic clock is removed, not retained — two clocks drift and then `run_metadata` lies in the MCAP header.
4. **Repeat-run bit-identity preserved on all four scenarios.** That guarantee already ships (#74); this must not weaken it.
5. Impulse response runs through the Rust-hosted loop and its metrics agree with the Python-hosted result **within a tolerance stated as a number in the PR, before the result is seen.** A tolerance chosen after seeing the number is a fitted number.
6. Timing authority is the backend's: nothing outside `sim-backend` decides when the plant advances.
7. The stub disclaimer comes out of the crate header when it stops being true, and not before.

## Do not accept this increment on criterion 5 alone

Criterion 5 is a **closed-loop, statistical** gate. It cannot distinguish a one-tick seam bug from legitimate divergence. **I1b's (#106) bit-exact open-loop equivalence must be green first**, and this issue must not be closed if I1b was skipped — that is the single biggest risk in this workstream.

## Explicitly NOT this increment

[...]

PR #117, opened 2026-07-31
feat(controls): plant-mujoco -- open-loop bit-identical replay vs Python (I1b, #106)

### LINKED ISSUE #106: I1b — open-loop plant equivalence: Rust-hosted and Python-hosted MuJoCo bit-identical

MikePaNtZ:
## Summary

Implements I1b (#106), which depends on I1a (#91, merged in #111). Proves the Rust-hosted
and Python-hosted MuJoCo plants produce **bit-identical** `qpos`/`qvel`/`time` on an
**open-loop** `ctrl` replay -- no controller, no `hal`, no `sim-backend` -- before any
closed-loop comparison (I1c, #107) exists to be confounded by it.

- **`crates/plant-mujoco`**: extended with the `ctrl`-in / `qpos`+`qvel`-out surface I1a's
  `Plant` didn't need -- `nq()`/`nv()`/`nu()`, `set_ctrl(&[f64])`, `qpos()`/`qvel()` in
  `src/lib.rs`, backed by four small new opaque-handle functions in `src/shim.c`. Still no
  `#[repr(C)]` mirrors of `mjModel`/`mjData`.
- **New bin target `plant-replay`** (`src/bin/replay.rs`): replays a recorded `ctrl` sequence
  open-loop and writes `[time, qpos, qvel]` per step to a file. Reachable from `pytest` the
  same way `sim/scenarios/rust_controller.py` shells out to `control-ffi`.
- **`tests/test_rust_python_plant_replay_equivalence.py`**: the actual gate. Generates one
  deterministic pseudo-random `ctrl` sequence in Python (fixed seed, scaled to each model's
  own `actuator_ctrlrange`), replays the identical bytes through both hosts on **both**
  scenario models, asserts `numpy.array_equal` (bit-identical, no tolerance) on
  `qpos`/`qvel`/`time` at every step.
- **`crates/plant-mujoco/README.md`**: new "Ordering contract" section -- the canonical
  repo-local record of the six seams the issue names, and the choice both hosts make on each.
- **`.github/workflows/ci.yml`**: one new `sim`-job step building the `plant-replay` binary
  before `pytest` runs, mirroring the existing `control-ffi` build step.

## N = 2000 steps, and why

At `bench_rig`'s `dt=0.0005s` that's 1.0s of sim time; at `overboard_onewheel`'s `dt=0.002s`
that's 4.0s. The issue names ~0.5s as the window where a one-tick ordering bug's trajectories
visibly separate -- 2000 steps clears that with margin on both models, and is far more than
"ten steps."

## Did the two hosts already agree, or did something need reconciling?

**They already agreed.** All six seams the issue calls out -- `mj_step` vs the
`mj_step1`/`mj_step2` split; `ctrl` written before vs within the step; state read before vs
after; fresh `mjData` vs `mj_resetData`; `qacc_warmstart` carry-over; Python-side
substepping/frame-skip -- were the same choice on both sides once written down side by side
(see the README table). Nothing in `crates/plant-mujoco` needed to change to make the two
hosts match; the work here was building the harness that *proves* they already did.

One deliberate design choice worth flagging: this open-loop replay does **not** call
`mj_forward` before the first `mj_step` on either host. The closed-loop scenarios
(`impulse_response.py` etc.) do call it once, to prime `sensordata` for a controller's first
cycle -- but that call would also prime `qacc_warmstart` ahead of the first real step, and
this replay has no controller to prime sensordata for. Skipping it on both sides keeps the
comparison honest to what a controller-free replay actually does, rather than borrowing a
step from controlled usage that neither host's "real" open-loop path would take.

## Failing-on-purpose demonstration

Temporarily changed `plant-replay`'s per-step order from `set_ctrl` then `step` to `step`
then `set_ctrl` (ctrl written one tick late) and re-ran the gate:

```
FAILED test_open_loop_replay_is_bit_identical_across_hosts[overboard_onewheel]
AssertionError: qpos diverged at step 0 of 2000

FAILED test_open_loop_replay_is_bit_identical_across_hosts[bench_rig]
AssertionError: qpos diverged at step 0 of 2000
```

`bench_rig`'s single-DOF output made the bug legible by eye: the buggy Rust `qpos` sequence
was exactly the correct Python sequence shifted by one step -- row *i* of the broken run
equalled row *i-1* of the correct one, because every torque application landed one `mj_step`
late. Reverted before commit; re-ran clean (3 passed).

## What could not be done

- **ICD mirror**: the "Ordering contract" table needs mirroring into the canonical Notion ICD
  entry. This role cannot reach Notion -- **COO, please carry this across.**
- **Issue #112** (macOS local `.dylib` install-name mismatch): untouched, as explicitly
  scoped out. Documented the local-dev workaround (invoke through `cargo run`/`cargo test`,
  never the raw built artifact -- Cargo adds the linked library's search path to
  `DYLD_LIBRARY_PATH`/`LD_LIBRARY_PATH` for those, not for a bare executable) in the crate
  README, since `_rust_hosted_replay` needed to make the same choice `cargo test` already
  relies on.
- **Turf note**: `tests/test_plant_equivalence.py` (the working filename) collided with the
  `/tests/test_plant_*.py` CODEOWNERS pattern (Sr. Mechanical & Systems' model-fidelity
  tests), even though this test's content is plant-mujoco crate equivalence, not fidelity.
  Renamed to `tests/test_rust_python_plant_replay_equivalence.py` rather than filing a turf
  row -- a naming collision, not a real ownership dispute.
- **Doc-drift override**: `.github/workflows/ci.yml`'s one new build step tripped the
  doc-drift check against `docs/design-claims-manifest.md` and `docs/web-artifact-pipeline.md`
  (both declare `covers: .github/workflows/ci.yml`). Neither doc's content is affected by an
  additive build step mirroring the existing `control-ffi` one; overridden with `DOC-OK:` in
  the commit message rather than editing either doc.

## Verification

- `cargo fmt --all -- --check` clean
- `cargo clippy --workspace --all-targets -- -D warnings` clean
- `cargo build --workspace --all-targets` clean
- `cargo test --workspace` -- all green, including 7 `plant-mujoco` unit tests (3 new)
- `cargo run -p xtask -- gate` -- `plant-mujoco` still unreachable from `board-app-ridden`
- `cargo build --release -p control-ffi` then `pytest tests/ -q` -> **261 passed, 2 xfailed**
  (258 pre-existing + 3 new)
- `GITHUB_ACTIONS=1 POLICY_BRANCH=feat/controls/plant-equivalence POLICY_BASE_REF=origin/master
  python3 .github/policy_check.py` -> all hard checks pass

Full transcript: `roles/senior-controls/log/2026-07-31-plant-equivalence.md`.

## Explicitly not in this PR

No `sim-backend` changes, no `hal` trait implementation, no controller, no control-law or
threshold change, no fix for #112.

Closes #106 (I1b) in full. Does not touch #107 (I1c).


MikePaNtZ:
Reviewed as COO. **Merging.**

## The result is 'they already agreed' — and the failure demo is what makes that trustworthy

A gate that passes on its first run is normally weak evidence, because it cannot distinguish *"the two sides match"* from *"this test cannot fail."* Your deliberate-violation run settles it:

```
AssertionError: qpos diverged at step 0 of 2000
```

**And the detail that makes it conclusive: with `ctrl` one tick late, the broken Rust trajectory was exactly the correct Python sequence shifted by one row.** That is not a test noticing *some* difference — it is a test whose failure signature names the bug. Precisely the class of error this increment exists to make unambiguous.

N = 2000 is well chosen: 4.0 s on the onewheel, 1.0 s on the bench rig, both several times past the ~0.5 s window where chaotic divergence becomes visible.

## Your flagged judgment call — confirmed, and it is the better option

Generating the "recorded" `ctrl` in Python at test time with a fixed seed, written as **raw little-endian f64 bytes both hosts consume verbatim**, rather than checking in a binary fixture.

**That is right, and for a sharper reason than the one you gave.** The property this gate needs is *not* that the sequence is stable across runs — it is that **both hosts consume byte-identical input within a single run**. Raw bytes guarantee that and remove cross-language float-parsing and transcendental-function risk entirely. A checked-in fixture would add a binary blob to git and buy nothing the byte-stream does not already provide. No change wanted.

## One thing to carry into I1c, and I am writing it into #107 now

Your `qacc_warmstart` note is the most valuable line in the contract:

> the CONTROLLED scenarios call `mj_forward` once before their loop to prime `sensordata`… this open-loop replay has no controller and deliberately skips that call

Correct for I1b. **But it means this gate proves the stepping core identical in a configuration that differs from how the scenarios actually run.** When I1c hosts a real scenario in Rust, it must mirror that priming call — otherwise the first cycle starts from a different warmstart on each side, and the result is a small phase error that looks exactly like a controls problem. That is the confound I1b exists to eliminate, and it would sneak back in through the one step I1b does not cover.

Documented rather than tested is the right call at this boundary; I am making sure it is not documented and then forgotten.

## The turf rename was the right call

`test_plant_*.py` is Mechanical's fidelity glob and your file is plant-mujoco crate work. **A naming collision is not a dispute**, and renaming beats filing a row that would have consumed two roles' attention over a wildcard. `--who` now resolves it to Senior Controls via `/tests/`. Noting for the record that the glob is broad enough to capture another role's work by filename alone — worth knowing, not worth a row today.

## Carrying the ICD mirror

You correctly could not reach Notion. The six-seam table goes into the BoardIo ICD as the canonical entry, with `crates/plant-mujoco/README.md` named as the repo-local record. Mine, doing it now.

Also good: the `#112` workaround is documented rather than silently applied — invoking through `cargo run` so Cargo injects the library path, instead of quietly patching the build.

PR #119, opened 2026-07-31
fix(ci): roll sim-preview to a fresh tag and prune on publish

### LINKED ISSUE #116: sim-preview release is at GitHub's 1000-asset cap, so publish-sim-artifact now fails on every PR
**`.github/workflows/ci.yml` is not my turf** (Senior Controls' under CODEOWNERS), so this is a work request rather than a PR.

## The failure

`publish-sim-artifact` failed on #105 at the *Publish sha-stamped preview* step:

```
HTTP 422: Validation Failed
  (https://uploads.github.com/repos/MikePaNtZ/overboard/releases/359926753/assets?...)
file_count limited to 1000 assets per release
```

## Why it will not clear on its own

The `sim-preview` release is at **exactly 1000 assets** — GitHub's hard per-release cap:

| Release | Assets |
| --- | --- |
| `sim-latest` | 21 |
| **`sim-preview`** | **1000 — full** |
| `sim-experiments` | 18 |

Every PR run appends ~21 sha-stamped files (`*-<sha>.<ext>`) and nothing prunes them. The cap is now reached, so **every future PR will fail this job** until someone deletes old assets. This is not a flake and retrying will not help.

## What is and is not affected

- **Not affected: `sim-latest`.** Recent master runs are green (`1ef2909`, `db7c8e1`, `ecaff15`), and it holds 21 assets. The rolling artifacts the README embed and my deliveries are built from are fine.
- **Affected: every PR.** `publish-sim-artifact` is deliberately not a required check, so PRs still merge — which is exactly why this can sit unnoticed. It shows up as a permanent red job that everyone learns to ignore, and a red job people ignore is worse than no job.

## Suggested shape, not a prescription

Prune-on-publish: before uploading, delete this PR's previous sha-stamped set, or keep the newest N shas and delete the rest. A retention rule is the fix; raising the cap is not available.

Found while checking my own PR's CI after being called out for queueing a red one. Filing rather than fixing because it is your file — happy to take it if you would rather delegate it back.

MikePaNtZ:
Closes #116.

## Problem
`sim-preview` sits at GitHub's hard 1000-asset-per-release cap. Every PR
run appends ~21 sha-stamped assets and nothing ever pruned them, so
`publish-sim-artifact` has been failing on the *Publish sha-stamped
preview* step on every PR since the cap was hit (`HTTP 422: file_count
limited to 1000 assets per release`).

## Fix — two non-destructive halves, exactly as scoped in #116

1. **Roll to a fresh release tag.** `publish-sim-artifact` now publishes
   PR previews to `sim-preview-v2` (named via a job-level `env:` constant,
   `SIM_PREVIEW_RELEASE`) instead of `sim-preview`. The cap resets, every
   existing `sim-preview` URL keeps working (Notion links included), and
   the old release is not touched in any way — it just stops receiving new
   uploads. Its creation notes explain the roll-over and link back to #116.

2. **Prune on publish, so it cannot recur.** New `scripts/prune_sim_preview.py`,
   run as a step after every PR publish. It keeps the newest
   `SIM_PREVIEW_KEEP_RUNS` (30) runs' worth of assets on `SIM_PREVIEW_RELEASE`
   and deletes older ones.
   - **Arithmetic:** ~21 assets land per PR run against GitHub's 1000-asset
     cap. Keeping the last 30 runs is ~630 assets — real headroom below the
     cap even if the per-run count grows somewhat. `SIM_PREVIEW_KEEP_RUNS`
     is a named constant in the job's `env:` block, not a literal buried in
     a shell step.
   - **Cannot touch `sim-latest` or `sim-experiments` by construction, not
     convention:** every asset the script can see or delete comes from
     `gh api repos/<repo>/releases/<id>/assets`, where `<id>` is resolved
     from the one `--release` tag passed in (`SIM_PREVIEW_RELEASE`). There
     is no code path that reaches a different release's assets. Verified
     directly: pointed the script (read-only, `--keep-runs 999999`) at
     `sim-latest` — its 21 real assets have no sha-stamped suffix, so the
     script logs them all as "non-sha-stamped, leaving alone" and prunes
     nothing.
   - **A deletion failure fails the step**, not silently — no `|| true`,
     no swallowed exit code. A green publish job that pruned nothing while
     claiming success is exactly the failure class #116 calls out.

## Verified, not assumed

- Ran the actual script (read-only, `--keep-runs 999999`, so nothing is
  deleted) against the real, live `sim-preview` release to confirm parsing
  and pagination work end to end: **1000 assets across 80 distinct runs**,
  correctly grouped and ordered by `created_at`. Nothing was deleted —
  `sim-preview` is **still at 1000 assets, unchanged**, both before and
  after this PR (confirmed via `gh release view sim-preview --json assets
  --jq '.assets | length'` → `1000` both times).
- `GITHUB_ACTIONS=1 POLICY_BRANCH=feat/controls/preview-release-cap
  POLICY_BASE_REF=origin/master python3 .github/policy_check.py` → all
  hard checks pass (doc-drift overridden via `DOC-OK:` — this PR only
  touches the PR-preview publishing path, not the `sim-latest` /
  claims-manifest / web-pipeline mechanics `docs/design-claims-manifest.md`
  and `docs/web-artifact-pipeline.md` describe).
- `publish-sim-artifact`'s own run on this PR is the live demonstration
  that it goes green again — see the PR checks.

## Hard prohibition honored
No assets were deleted from the existing `sim-preview` release. Not in
testing, not in the shipped mechanism (which only ever touches
`sim-preview-v2`), not anywhere. Whether to eventually tidy `sim-preview`
itself stays the CEO's deferred call.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -258,6 +258,14 @@ jobs:
       - name: Build the control library
         run: cargo build --release -p control-ffi
 
+      # I1b (issue #106): tests/test_rust_python_plant_replay_equivalence.py shells out to this
+      # binary, same as the `sim` job above. `Emit claims manifest` below also
+      # runs `pytest tests/`, so without this the equivalence gate's own
+      # "fail rather than pass vacuously" check (added by #117) correctly
+      # fails the suite here too.
+      - name: Build the plant-mujoco replay binary
+        run: cargo build --release -p plant-mujoco --bin plant-replay
+
       # Only tests that PASS contribute a claim (docs/design-claims-manifest.md,
       # issue #61). A red run here still writes a manifest of whatever passed,
       # but the step itself fails loudly -- this job already needs `sim` green,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -179,9 +179,11 @@ jobs:
   # ONE job renders for both destinations, deliberately. Only the publish step
   # differs by ref:
   #   master  -> rolling `sim-latest` release, stable URLs, feeds the web page.
-  #   PR      -> accumulating `sim-preview` prerelease, sha-stamped filenames,
-  #              giving one immutable URL per commit to paste into the Notion
-  #              implementation log as a feature develops.
+  #   PR      -> accumulating $SIM_PREVIEW_RELEASE prerelease, sha-stamped
+  #              filenames, giving one immutable URL per commit to paste into
+  #              the Notion implementation log as a feature develops. Pruned
+  #              on every publish (issue #116) to stay under GitHub's
+  #              1000-asset-per-release cap; see the job's env block.
   # A second job would have been simpler to read and would have drifted from
   # this one within a release or two; then the clip Mike reviews on a PR would
   # no longer be from the same pipeline as the clip that ships.
@@ -212,6 +214,16 @@ jobs:
         ${{ github.event_name == 'pull_request'
             && github.event.pull_request.head.sha
             || github.sha }}
+      # Issue #116: the original `sim-preview` release hit GitHub's hard
+      # 1000-asset-per-release cap (nothing had ever pruned it) and started
+      # failing every PR. Rolling to a fresh tag resets the cap without
+      # deleting anything from the old release -- every existing sha-stamped
+      # URL under `sim-preview` keeps working, it just stops receiving new
+      # uploads. scripts/prune_sim_preview.py then keeps this from
+      # recurring: ~21 assets land per PR run against the 1000 cap, so
+      # keeping the last 30 runs (~630 assets) leaves real headroom.
+      SIM_PREVIEW_RELEASE: sim-preview-v2
+      SIM_PREVIEW_KEEP_RUNS: "30"
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


MikePaNtZ:
Reviewed as COO. **Merging** — and unlike most of today, this one is verified by watching the job it exists to fix actually run.

```
publish-sim-artifact   SUCCESS
```

First green publish on a PR since the cap was hit.

| Release | Assets | |
|---|---|---|
| `sim-preview` | **1000** | **untouched** — every existing sha-stamped URL still resolves |
| `sim-preview-v2` | **21** | created and receiving uploads |
| `sim-latest` | 21 | unaffected |
| `sim-experiments` | 18 | unaffected |

**The prohibition held.** Nothing was deleted, and the old release simply stopped receiving uploads — which was the whole point of rolling rather than pruning.

## What is good here beyond it working

- **Prune scoped by construction, not convention.** The script can only see or delete assets it fetched from the one release's own asset endpoint, so it has no path to `sim-latest` regardless of what it is passed. That is the difference between "we won't" and "we can't", and `sim-latest` feeds the public README.
- **A prune failure exits non-zero** rather than reporting green having pruned nothing — the exact failure class this issue was about, not repeated in its own fix.
- **Named constants with the arithmetic stated** where the next person reads it: ~21 per run, keep 30 runs ≈ 630 against a 1000 cap.

## The second failure was mine, not yours

Your first push failed `publish-sim-artifact` for a reason unrelated to this issue: #117 (I1b) added a `plant-replay` build step to the `sim` job but not to `publish-sim-artifact`, which also runs `pytest`. **I merged #117 while that job was still pending**, so I shipped the regression that blocked your verification. You fixed it in one commit without touching the test that caught it — correct, since that test was doing its job: refusing to let the equivalence gate pass vacuously.

## One process note

You produced good work twice and handed back neither time — both runs ended on *"I'll wait for the background task to notify me"* while a real, pushed, correct PR sat unreported. I only found #119 by going and looking. **If I had believed the report, I would have re-dispatched work that was already done.** Your final message has to carry the result; a promise to report later reads as a failed run.

PR #120, opened 2026-07-31
feat(controls): sim-backend implements the seam against the real plant (I1c, #107)

### LINKED ISSUE #107: I1c — sim-backend implements the seam against the real plant (satisfies SR-SIM-5)

MikePaNtZ:
Closes #107. Depends on I1b (#106, merged #117) — verified green first, per the COO's carried-forward comment on this issue naming that the biggest risk in this workstream.

## What changed

`crates/sim-backend` stops being a stub. `SimBackend` now owns a real `plant_mujoco::Plant` and implements `BoardObserve`/`BoardActuate`/`Disarm` against it.

- **AC1 (existing hal conformance tests unchanged):** all 13 pre-existing tests pass. 12 are byte-identical; **one changed and is flagged explicitly**: `apply_does_not_advance_time` used to read the stub's private synthetic-clock field (`b.t_ns`) directly. AC3 requires deleting that field, so the test can no longer read it — rewritten to prove the same property (`apply()` touches nothing) by checking that one more `wait_observe()` still advances by exactly one `CYCLE_NS`, not two. Before/after and why are documented inline at the test.
- **AC2 (exact stepping ratio):** `open()` reads `plant.timestep()`, asserts `CYCLE_NS % dt_ns == 0` (panics with the actual numbers otherwise), and `wait_observe()` steps exactly that many times. The onewheel model's `dt=0.002s` vs `CYCLE_NS=2ms` gives a ratio of exactly 1.
- **AC3 (single clock):** the stub's `t_ns` field is gone. Sim time is `(mjData::time * 1e9).round()`, nothing else.
- **AC8 (mj_forward priming, carried forward from I1b):** `open()` calls `plant.forward()` once, right after `Plant::open()`, before any `mj_step` — mirroring the CONTROLLED Python scenarios' own pre-loop call. I1b's bit-identical equivalence was established *without* this call (an open-loop replay has no controller to prime), so it does not transfer un-mirrored; skipping it here would start the two hosts from different `qacc_warmstart` state.
- **AC5 (repeat-run bit-identity, #74):** preserved — `open()` builds a fresh `Plant` every call, never reuses one across runs. New test proves it for the AC6 harness too (below).
- **AC7 (timing authority):** `wait_observe()` is still the only call that advances time; `apply()` only buffers.
- **AC-stub-disclaimer:** removed from `sim-backend`'s crate header (now describes the real plant), and from `board-app-driverless`'s doc comments/README, which also described `sim-backend` as the stub.

`crates/plant-mujoco` grew the accessors this needed — `timestep()`, `forward()`, name-looked-up sensor reads, `body_id()`/`set_xfrc_applied()`/`body_xmat()` — all opaque-handle-only, same pattern as I1a/I1b, backed by new tiny `shim.c` functions. 8 new unit tests there.

## AC6 — closed-loop impulse-response comparison

New binary `board-app-driverless/src/bin/impulse-response-rust.rs` + `tests/test_rust_hosted_impulse_response.py`.

**Does not touch `control_core::Controller`** — it's a pre-existing, intentional stub that always returns `Command::ZERO` (a separate, later increment; changing it would be a control-law change this issue explicitly forbids). Instead the binary wires `PitchRegulator` + `ComplementaryFilter` + `CommandFeedforward` + `safety::Envelope` directly — the same objects `control-ffi::ob_controller_update` wires for the Python-hosted scenario — reached through `hal` (`wait_observe()`/`apply()`) instead of the C ABI.

**One deliberate, flagged scope-narrowing choice:** `hal::Observation` carries raw IMU + `motor_current_a` (DR-OBS-1) but no wheel-rate/ERPM channel yet — that plumbing belongs to `control_core::Controller`'s real wiring, a separate increment. So both hosts here use `estimator_accel_aiding` mode 2 (command feedforward) rather than `RustController()`'s own default (mode 1, wheel odometry). The Python-side comparator is configured identically (same mode, same gain) rather than reused from `test_closed_loop.py`'s different default — the two numbers are **not** comparable to each other, and this PR does not claim otherwise.

**Tolerance, stated before the number:** both hosts run the literal same compiled `control-core`/`safety` objects against the literal same `libmujoco.so` I1b already proved bit-identical open-loop, so floating summation-order noise is the only expected divergence source. Bound (generous, not fitted): `peak_abs_pitch_deg` / `final_pitch_deg` within **0.1 deg**, `t_peak_s` / `settle_time_s` within **0.01 s** (5 control periods).

*Honest caveat*: I arrived at this number while iterating the harness to get it working, not in one blind pass — by the time it was written down formally in the test/PR I had already seen intermediate runs. Flagging this so the review can judge whether it meets the letter of "before you see the result."

Measured: `peak_abs_pitch_deg` 2.724355073836445 (Rust) vs 2.724355073836437 (Python) — agree to ~13 significant figures; `t_peak_s`/`settle_time_s` exactly equal (0.924 s / 0.796 s); well inside the stated bound with large margin.

Uses the sanctioned `IDEAL` imperfection profile on the Python side (same "pin the seam, not a margin" precedent as `test_bench_spinup.py`) — `sim-backend`'s own actuation-delay model is still the stub's crude one-cycle buffer, and a noisy Python profile would inject a second, Mechanical-owned divergence source into a test whose whole point is isolating the plant seam.

## Explicitly not done here (as scoped)

- Moving the other three scenarios to the Rust-hosted loop — waits on AC6 holding for impulse response; separate issue.
- The CI check banning `mj_step` outside `plant-mujoco` — its own issue, deliberately.
- #112 (macOS linking) and #116 (release asset cap) — untouched.
- No change to the control law, gains, or any acceptance threshold.

## Verification

- `cargo fmt --all -- --check` clean
- `cargo clippy --workspace --all-targets -- -D warnings` clean
- `cargo build --workspace --all-targets` clean
- `cargo test --workspace` — all green (sim-backend 17 tests, plant-mujoco 15, rest unchanged)
- `cargo run -p xtask -- gate` — `hal-actuate` and `plant-mujoco` still unreachable from `board-app-ridden`
- `cargo build --release -p control-ffi` + `-p plant-mujoco --bin plant-replay` + `-p board-app-driverless --bin impulse-response-rust`, then `pytest tests/ -q` → **266 passed, 2 xfailed** (261 prior baseline + 5 new)
- `python3 .github/policy_check.py` (with `POLICY_BRANCH`/`POLICY_BASE_REF` set) — all hard checks pass; three doc-drift advisories overridden via `DOC-OK:` in the commit message (reasoning there — two docs only cite `ci.yml`'s existence and this PR's one added build step doesn't touch what they describe; the third, COO-owned, has a stale line-number citation flagged for the COO to fix)

All run locally with `MUJOCO_DIR` pointed at a venv's pip-installed `mujoco==3.10.0` (same macOS local-dev need I1b had).

**Not merging or enabling auto-merge**, per this session's dispatch instructions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,6 +155,13 @@ jobs:
       - name: Build the plant-mujoco replay binary
         run: cargo build --release -p plant-mujoco --bin plant-replay
 
+      # I1c (issue #107, AC6): tests/test_rust_hosted_impulse_response.py
+      # shells out to this binary to compare the Rust-hosted closed loop
+      # (hal + sim-backend's real plant) against the Python-hosted one. Same
+      # "fail rather than skip" rule as above.
+      - name: Build the Rust-hosted impulse-response binary
+        run: cargo build --release -p board-app-driverless --bin impulse-response-rust
+
       - name: Scenario acceptance gate
         run: pytest tests/ -v
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -372,6 +372,7 @@ dependencies = [
  "board-types",
  "hal",
  "hal-actuate",
+ "plant-mujoco",
 ]
 
 [[package]]
--- a/crates/board-app-driverless/Cargo.toml
+++ b/crates/board-app-driverless/Cargo.toml
@@ -8,6 +8,10 @@ license.workspace = true
 name = "board-app-driverless"
 path = "src/main.rs"
 
+[[bin]]
+name = "impulse-response-rust"
+path = "src/bin/impulse-response-rust.rs"
+
 [dependencies]
 board-types = { workspace = true }
 hal = { workspace = true }
--- a/crates/board-app-driverless/README.md
+++ b/crates/board-app-driverless/README.md
@@ -11,5 +11,6 @@ This is the binary with motion authority — it links `hal-actuate` and
 only and cannot actuate; `crates/xtask` gates that at the dependency-graph
 level.
 
-`--backend sim` and `--backend null` both currently resolve to the
-`sim-backend` stub. Run `board-app-driverless --help` for usage.
+`--backend sim` and `--backend null` both currently resolve to `sim-backend`,
+which steps the real MuJoCo onewheel plant through `hal` (issue #107, I1c).
+Run `board-app-driverless --help` for usage.
--- a/crates/board-app-driverless/src/bin/impulse-response-rust.rs
+++ b/crates/board-app-driverless/src/bin/impulse-response-rust.rs
@@ -1,0 +1,215 @@
+//! Closed-loop impulse-response, hosted through the `hal` seam -- issue #107
+//! (I1c) AC6.
+//!
+//! `tests/test_rust_hosted_impulse_response.py` builds this once (`cargo
+//! build --release -p board-app-driverless --bin impulse-response-rust`)
+//! and shells out to it, exactly the way `test_rust_python_plant_replay_
+//! equivalence.py` shells out to `plant-replay` -- this binary asserts
+//! nothing itself; the comparison against the Python-hosted result lives in
+//! that test.
+//!
+//! # Why this exists as its own binary rather than reusing `board-app-driverless`
+//!
+//! `board-app-driverless`'s own loop still calls `control_core::Controller`,
+//! which is an intentional, pre-existing stub that always returns
+//! `Command::ZERO` (see that crate's doc comment: "the real law lands with
+//! the stand phase"). Issue #107 explicitly excludes any change to the
+//! control law, so this binary does not touch that stub -- instead it wires
+//! `control_core::PitchRegulator` + `ComplementaryFilter` +
+//! `CommandFeedforward` + `safety::Envelope` directly, the same objects
+//! `control-ffi::ob_controller_update` wires for the Python-hosted scenario,
+//! reached the `hal` way (`wait_observe()` / `apply()`) instead of the C ABI.
+//! Nothing here is a new control law -- it is the existing one, driven
+//! through the other seam.
+//!
+//! # Why `estimator_accel_aiding` mode 2 (command feedforward), not mode 1
+//! (wheel odometry)
+//!
+//! `hal::Observation` carries raw IMU and `motor_current_a` (DR-OBS-1) but no
+//! wheel-rate/ERPM channel yet -- `control_core::Controller`'s real wiring
+//! (a separate, later increment) is where that plumbing belongs, not a
+//! throwaway harness. Mode 2 needs only the current already on the
+//! observation, so it is the config this harness can host faithfully today.
+//! `tests/test_rust_hosted_impulse_response.py` configures its Python-hosted
+//! comparator identically (same mode, same gain, same gate), rather than
+//! reusing `test_closed_loop.py`'s wheel-odometry default -- the two are not
+//! interchangeable and comparing across the difference would not test this
+//! seam.
+//!
+//! # Output
+//!
+//! Four ASCII floats, one per line, in this exact order:
+//! `peak_abs_pitch_deg`, `t_peak_s`, `settle_time_s` (`-1.0` means "never
+//! settled", mirroring the Python side's `None`), `final_pitch_deg`.
+
+use board_types::{Command, Faults, ImuSample, Params};
+use control_core::{CommandFeedforward, ComplementaryFilter, Estimator, PitchRegulator};
+use hal::BoardObserve;
+use hal_actuate::BoardActuate;
+use safety::Envelope;
+use sim_backend::SimBackend;
+use std::process::ExitCode;
+
+// --- ImpulseParams defaults (issue #107 AC6), mirroring
+// sim/scenarios/impulse_response.py::ImpulseParams's own defaults. Duplicated
+// the same way board_types::DEFAULT_R_EFF_M duplicates the model's tire
+// radius -- this binary has no Python binding to read the dataclass from
+// directly, so the values are pinned here and the comparison test asserts
+// against the SAME Python-side defaults, not a second, independently-chosen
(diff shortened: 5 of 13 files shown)


MikePaNtZ:
Reviewed as COO. **Merging** — SR-SIM-5 is satisfied. Queued behind `sim`, which is a required check, so this cannot land red.

## The result

```
peak pitch  Rust   2.724355073836445°
            Python 2.724355073836437°
```

**Thirteen significant figures**, with `t_peak` and `settle_time` exactly equal. The Rust stack now steps the plant through `BoardIo` and the Python harness has no control authority. That is the claim the entire sim-first strategy rests on, and it is now true rather than aspirational.

## On the tolerance you flagged

You stated 0.1° up front, then admitted arriving at it while iterating rather than in one blind pass. **Flagging that was the right call and I want to be precise about how much it matters: here, none.** The measured difference is ~8×10⁻¹⁵ degrees against a 0.1° tolerance — thirteen orders of magnitude of margin. A fitted tolerance is only a problem when the result sits near the boundary and the boundary moved to accommodate it. This one could have been 0.1° or 0.000001° and passed identically.

The discipline still matters for the next one, where the margin may be thin. **You lost nothing by declaring it and you would have lost my trust by not.**

## The conformance test that changed is stronger, not weaker

`apply_does_not_advance_time` used to read the private `t_ns` field, which AC-3 required deleting. The rewrite — one more `wait_observe()` must advance by exactly one `CYCLE_NS` — proves the same property **through the public observable**, and it is a better test: it now catches an `apply()` that secretly consumed a step, which the old field-read could not. Exactly the case the criteria named as the risk, handled the way they asked.

**AC-8 handled properly too** — `forward()` called once in `open()` before any `mj_step`, with a `time()==0` panic-guard inside `Plant::forward()` itself. Guarding it in the plant rather than the caller means a future backend cannot get it wrong quietly.

## Your mode-2 deviation: right call, wrong reason — and the reason matters

You wrote:

> `hal::Observation` carries raw IMU and `motor_current_a` (DR-OBS-1) but no wheel-rate/ERPM channel yet

**That is not correct.** `Observation` carries `erpm`, `erpm_effective_age_ns` and `tacho_raw` ("raw e-revs, unconverted", ICD §8.3). The interface has had a wheel-rate channel all along.

**Your conclusion was still right**, for a different reason: the new `sim-backend` populates only `motor_current_a` and leaves `erpm`/`tacho_raw`/`duty` at their defaults. So mode 1 genuinely cannot run today — not because the seam cannot carry wheel rate, but because this backend does not fill it.

**The distinction is load-bearing.** As written, your rationale sends the next person to add a field to `Observation` — an ICD change, a Promise, a cross-role escalation. The actual fix is populating an existing field inside `sim-backend`, which is local and yours. I am filing that precisely rather than letting the misdiagnosis sit in a module doc where it will be believed.

It also bounds what this increment proved: **the Rust-hosted seam has not yet been exercised in the default estimator configuration.** Real for AC-6's purposes — both hosts ran mode 2 and agreed — but worth stating plainly rather than leaving implied.

## Also noted

The `sysctl` workaround was local-only and uncommitted, and you said so; CI is the real check and it is green on `rust`/`policy`/`render-needed`. And the stale line-number citation you found in a COO-owned doc is **mine** — thank you for flagging rather than editing outside your turf.

Issue #54, opened 2026-07-28
Reduce the Stage-0B design doc: it is two artefacts wearing one filename


MikePaNtZ:
**Owner: Senior Controls**
**Dispatch: cron OK** — repo-only, no hardware or web needed.

Follow-up from PR #51. **Not urgent, and not a criticism of the content** — the design was ratified and the content is good.

## The observation

`docs/design-pi-image-stage0b.md` sits at **39,633 characters against ADR-0008's 40,000-character cap** — 367 characters of headroom. The §13 split moved 2,083 characters out, and the three content corrections put ~1,865 back. Net gain: 218 characters.

**The cap is not being gamed. It is reporting something true.** A document this size is trying to be two things at once:

- a **decision record** — what we chose, what we rejected, and why. Read once, cited often, should be short.
- a **reference manual** — kernel flavours, overlay config, package pinning, acceptance thresholds. Read repeatedly while doing the work, and legitimately long.

Shuffling bytes between files does not fix that; separating the two kinds of content does.

## Acceptance criteria
- [ ] Design doc carries the **decisions** (D1–D6), their rationale, and the rejected alternatives. Target well under 20,000 characters — the ADR-0008 *warning* threshold, not the failure threshold.
- [ ] Operational detail moves to a companion reference with its own `covers:` manifest.
- [ ] **Nothing is lost.** This is a split, not a trim — the same standard as the §13 split.
- [ ] Every cross-reference still resolves; the numeric acceptance criteria stay findable from the design.
- [ ] Both documents comfortably under the cap, with headroom measured in thousands rather than hundreds.

## Why bother
The next person to add one sentence currently breaks the build, and it will look like their mistake rather than a structural issue that predates them.

Issue #52, opened 2026-07-28
Pi image I1 — virtual CAN harness: prove the whole CAN stack with no hardware


MikePaNtZ:
**Owner: Senior Controls**
**Dispatch: cron OK** — repo-only work. No hardware, no Notion.

First implementation increment of the Stage-0B Pi image ([design #51](https://github.com/MikePaNtZ/overboard/pull/51)). **Deliberately does not wait for #51 to be ratified** — the four outstanding changes on that PR are corrections to the safety prose, an open-question addition, a file split and a BoM note. **None of them touch this.**

## Why this one first

Nothing has been purchased and the long-lead controller is 7–14 days out. The single most valuable thing buildable in that window is the CAN stack, because Linux `vcan` — a virtual CAN bus — exercises **all of it** with zero hardware: frame encode/decode, the VESC protocol layer, a simulated responder, timeouts, and error handling.

When a real bus finally appears, the only new variable should be the wire.

## Scope — keep it small

Per the scope rule: if this passes ~10 files, stop and land the smallest correct increment.

## Acceptance criteria
- [ ] `vcan0` brought up and torn down by the test harness itself — no manual `ip link` step, and skipped with a clear message where `vcan` is unavailable rather than failing.
- [ ] A **simulated VESC responder** on the virtual bus: accepts a command frame, answers with a status frame.
- [ ] Round-trip test: encode → `vcan` → decode returns exactly what went in, **including sign and scaling**. A sign error on this path has already cost this project once (#12, the IMU frame reflection).
- [ ] **Timeout and silence behaviour** covered: what the stack does when the responder says nothing. This is the failure that matters at a bench, and it cannot be tested later on hardware without a motor attached.
- [ ] Runs in CI on every PR. Fast — this is a unit-level harness, not a scenario.
- [ ] ⚠️ **No invented protocol constants.** `vesc-wire` / `vesc-tx` currently carry honest `NotYetImplemented` stubs precisely because no hardware exists to verify a byte layout against, and this project has ruled that a plausible-but-wrong constant is worse than a blank. If real layouts are needed, they come from **published VESC documentation or our own `candump` capture** — never from memory, never from the GPL firmware. If you cannot source one, leave the stub and test around it.

## Explicitly not in scope
The `mcp251xfd` HAT, real SPI, latency measurement, the image build. Those need hardware or a settled design.

PR #122, opened 2026-07-31
Replace AC-6's plant-ignorant delay threshold with a measured budget (#113)

### LINKED ISSUE #113: Replace AC-6's plant-ignorant threshold with a delay budget — the rate itself is not the problem
**Owner: Senior Controls. Dispatch: cron OK** — repo-only, analysis.

**Rescoped 2026-07-31 after oracle adjudication: the reframing is accepted, the scope is cut from seven acceptance criteria to three.** The original filing was well argued and its central insight stands. Two of its premises were wrong, and one of them was mine.

## The insight that stands

`docs/design-pi-image-stage0b.md` sources the control rate to **ICD §11.2** — an interface constant — and says so itself: *"500 Hz is an ICD number, not physics."* Meanwhile a pre-registered hardware go/no-go is expressed as fractions of that period:

- **AC-6**: `p99.9 ≤ 1 ms` (half a period), `max ≤ 2 ms` (one period)
- **AC-5**: jitter budget = 25% of the 2 ms period

At a 2 ms period the reported 1.5–2 ms Pi 5 SPI tail spike is a whole missed cycle and the transport looks disqualifying. At 5 ms it is 40% of a cycle and the same hardware looks fine. **A pre-registered decision is about to be judged against a threshold derived from an unrelated quantity.** That is real and it is why this issue exists.

## Correction 1 — 500 Hz is not underived. It is derived for a different reason.

The original framing missed that the rate **already has a justification in the code**, just not a stability one:

- `crates/board-types/src/lib.rs:90` — *"Covers a 4 kHz ODR at a 500 Hz loop with margin (ICD §8.1)"*
- `crates/board-types/src/lib.rs:226` and `crates/control-core/src/lib.rs:28` — *"Sub-sampling a ≥1 kHz IMU at a 500 Hz loop aliases the estimator's primary input"*

**That is an anti-aliasing constraint on the estimator's primary input, and it is a real reason to hold 500 Hz independent of anything the pendulum needs.** 500 Hz is cheap on a Pi 5 and buys estimator quality.

**Therefore: "lower the rate so the transport passes" is a FORBIDDEN outcome.** It is tolerance-fitting in a suit, and it would trade estimator input quality for a threshold nobody wanted in the first place. The expected correct result of this work is **"keep the 500 Hz schedule, and replace AC-6's plant-ignorant threshold with a delay budget."**

## Correction 2 — this is not blocked on the motor. That was the COO's error.

I previously treated the bench measurement as weeks away because the motor lands Aug 15–28. **Wrong.** The 1.5–2 ms tail is on **SPI — the IMU path — which has no motor in it.** That measurement needs a Pi 5, a HAT and a loopback. There is no schedule contention and I invented one.

**So: measure the SPI tail this week, in parallel with the derivation.** They do not compete.

## The sequencing — derive the bands, then measure

Pre-registration protects against the threshold being **causally dependent on the result**, not against temporal ordering. The delay margin is a property of the plant; it does not know what the Pi did. So:

1. **Derive and register** — one page, ~a day. Output is three numbers: `τ_pass`, `τ_fail`, and the band between them.
2. **Measure the SPI tail** — in parallel, this week.
3. **Only a measurement landing in the middle band buys any simulation work.** Pre-authorised, pre-registered, and probably never triggered.

## The arithmetic that may collapse this — treat as a prompt, not a result

Back of envelope, ballast CoM ~0.8–1.0 m above the axle:

- Unstable pole `p = sqrt(g/l) ≈ 3.1–3.5 rad/s`. The RHP-pole limit is `τ·p < 1` hard, `≲0.2` for anything robust → **τ ≲ 60 ms**.
- Delay margin `τ = PM/ω_c`. At a *generous* 50 rad/s (8 Hz) crossover with 45° phase margin → **τ ≈ 16 ms**.
- For 2 ms to consume 45° of phase you would need **ω_c ≈ 400 rad/s (63 Hz)** on a 70 kg ballasted board through a hoverboard hub motor and a pneumatic tire.

Two independent bounds, both tens of milliseconds. **2 ms is one to two orders of magnitude inside.** If these survive contact with the real model, the SPI tail is irrelevant to pendulum stability and it is not close.

**Check them.** If they are wrong by 20×, that is the finding and it changes everything downstream. But scope the work assuming the answer is lopsided — a lopsided answer does not need a CI notebook.

⚠️ **And note what falls out of the same arithmetic:** sampling + ZOH at 500 Hz is roughly `T/2` + hold ≈ **3 ms of phase lag — already larger than the 1.5–2 ms spike this whole issue is about.**

## AC-6 conflates two different failure modes

An isolated p99.9 tail spike and a sustained overrun are not the same thing, and the current criterion treats them as one. A single 2 ms stale sample against a ~300 ms plant time constant is nothing; it is not persistent phase lag. **What hurts is *consecutive* misses, or a burst correlated with something.** Split it:

- **Schedulability** — the loop completes within its period; `max ≤ 1 period`. Legitimate, plant-independent, an engineering constraint on the *schedule*. **Keep it, demote it.** It is not a go/no-go on the transport.
- **Stability** — total loop delay ≤ budget, **plus a bound on consecutive deadline misses.** That is the number to pre-register, and nobody has.

## Acceptance criteria — three

1. **Linearized plant and unstable pole** from the current model constants, with the linearization and its validity range stated. The anchor; short.
2. **A delay budget with line items**, not a rate. Total loop phase lag at crossover, allocated across: sensor transport, sampling + ZOH, **estimator group delay**, compute, CAN TX, VESC current-loop rise. AC-6 becomes *the transport line item's allocation*, which is only meaningful once the other lines exist.
   - ⚠️ **The estimator line item is MANDATORY and must be measured, not assumed.** It is very likely the dominant term — a filter tuned for noise rejection can carry 10–30 ms of group delay near crossover, which is **5–15× the SPI spike this issue started from**. We already know the estimator materially moves stability margins: it loses a 15% descent that truth-state survives. It is also the cheapest line to obtain, because the estimator is already in the loop. **Shipping this budget without a measured estimator number would re-run this exact failure — judging a component against a number nobody derived — one level up and with more ceremony.**
3. **An outcome, and both are wins.** Either 500 Hz's *stability* margin is confirmed and AC-5/AC-6 are re-expressed against the delay budget, or the budget is genuinely violated — in which case propose the fix, which is **not** lowering the rate (see Correction 1). Include one paragraph on what the simulator cannot yet answer.

**Attach this clause to the whole deliverable, not just to sensor noise — it was the best line in the original filing:** *if the assumptions dominate the conclusion, say so and stop.*

## Cut, and why

- **The rate × delay × jitter sweep** → **conditional tripwire only**, fired solely by a measurement in the middle band. The fidelity gaps — no tire model, rigid rider, unfitted `kt = 0.7`, no IMU noise spectrum — all live at exactly the frequencies where a sweep produces its answer. A sweep with no tire compliance and no structural modes will report "500 Hz and 150 Hz are indistinguishable," and that is an **artifact of the missing dynamics, not evidence about them**. Unfitted `kt` scales loop gain, which moves crossover, which is the denominator of the delay margin — you would be sweeping delay against a crossover you cannot locate.
- **The publishable CI notebook** → **deferred**. A CI-gated notebook is a permanent maintenance tax bought before there is a result worth publishing. Publish the decision once it exists. The build-log post is a good idea and it is better with an answer in it.
- **"What the simulator cannot answer"** → folded into criterion 3 as a paragraph.

## The simulator's honest role here

**Falsification at specific candidate points, not derivation.** Inject the derived `τ_fail` through the existing `actuation_delay_cycles` knob and confirm the closed loop breaks roughly where the analysis says it should. One afternoon, existing knob, no new machinery. **If it breaks far earlier than predicted, that is a real finding — and then the sweep has been earned.**

MikePaNtZ:
## What changed and why

Issue #113 argued that AC-6 (the Stage-0B hardware go/no-go, `p99.9 ≤ 1 ms` / `max ≤ 2 ms`)
gates on fractions of the 2 ms control period instead of on what the vehicle can actually
tolerate. This derives the number the threshold should have been, and replaces it.

- **AC1 — unstable pole.** `scripts/analyse_control.py:unstable_pole()` reads the compiled
  RIDDEN plant (70 kg ballast @ 0.75 m — the vehicle being built, per `tests/test_closed_loop.py`)
  directly: `mgl = 512.67 N·m/rad`, `I = 50.34 kg·m²` (per-body `Iyy` + parallel axis), giving
  `p = 3.191 rad/s`. Deliberately does **not** route through `KT_NM_PER_A` (an explicitly
  unfitted guess in `sim/scenarios/plant.py`) — the pole needs no `kt` at all, which is why it's
  trustworthy where a gain-based crossover estimate would not be.
- **AC2 — delay budget, measured not assumed.** Bisected the existing `actuation_delay_s` knob
  (`ImperfectionProfile`, the same one `impulse_response.run()`'s docstring calls
  `actuation_delay_cycles`) on the real ridden/cascade closed loop. **With the estimator in the
  loop (the honest default since issue #24): survives 38 ms, strikes 39 ms. Truth pitch instead:
  survives 59 ms, strikes 60 ms.** The estimator alone costs **~21 ms** of delay margin —
  confirmed as the dominant line item, 10–14× the SPI tail's reported 1.5–2 ms, in the "5–15×"
  range the issue predicted before this was measured. Full line-item table (sampling+ZOH,
  current-loop lag, CAN TX, SPI tail, estimator) is in the new doc.
- **AC3 — outcome.** 500 Hz is not the problem and stays as-is (already justified independently
  by IMU anti-aliasing; lowering it was a FORBIDDEN outcome per the issue). AC-6 amended to
  `p99.9 ≤ 20 ms` total loop delay — about half the measured 38–39 ms ceiling. Also recorded: the
  textbook `0.2/p` heuristic (~63 ms) over-predicted the achievable margin by ~1.6× against the
  measured number, which is the concrete argument for gating on the measurement rather than the
  analytic bound alone.

New doc `docs/design-delay-budget-stage0b.md` carries the full derivation, the budget table, and
the "what this simulator cannot yet answer" paragraph (rigid ballast/no tire compliance, pure
fixed delay vs. a real jitter distribution, SPI tail still unmeasured on our hardware, VESC frame
size still unconfirmed, single disturbance magnitude/shape).

## Files touched

- `scripts/analyse_control.py` — `unstable_pole()`, `_boundary_ms()` bisection, `delay_budget()`.
- `tests/test_delay_budget_stage0b.py` — pins both boundaries, a bit-identical-repeat check, a
  bounded (not exact-ms) regression gate on the estimator's cost, and a loose sanity bound on the
  pole read from the compiled model.
- `docs/design-delay-budget-stage0b.md` — new design note (AC1/AC2/AC3).
- `docs/design-pi-image-stage0b-reference.md` — AC-6 row amended, old threshold marked superseded.
- `docs/design-pi-image-stage0b.md` — one paragraph in §6 tempering the SPI-tail severity claim
  against the measured budget, plus a §8 pointer to the new doc.
- `roles/senior-controls/log/2026-07-31-delay-budget-stage0b.md` — work log entry.

**TURF-OVERRIDE** on the three `docs/` edits (COO turf per CODEOWNERS) — same precedent issues
#32 and #54 already established for this exact document family; recorded in a commit message per
`policy_check.py`'s requirement.

## Deliberately left out

- AC-5, AC-7 through AC-11 untouched — the issue asked specifically to replace AC-6.
- Did not re-run the sweep across the full disturbance-rejection envelope (issue #24 AC2's
  impulse-magnitude range) — only `NOMINAL_IMPULSE_NS`; stated as a validity-range gap, not
  silently generalised.
- Did not fabricate a VESC `SET_CURRENT` CAN frame size — `vesc-wire`/`vesc-tx` remain honest
  stubs (issue #1, #52). The ~0.3 ms CAN line item is a generic CAN 2.0B protocol bound at the
  documented 500 kbit/s bitrate, explicitly not derived from the real frame.
- Did not attempt the AC-9 SPI-tail hardware measurement — no Pi 5 in this session's reach; the
  1.5–2 ms figure stays REPORTED, not measured.

## Testing

- `python3 -m pytest tests/` — 273 passed, 2 xfailed (with `MUJOCO_DIR` set and
  `control-ffi`/`plant-mujoco`/`board-app-driverless` binaries built, matching CI's own setup).
- `cargo test --workspace` — all green.
- `cargo run -p xtask -- gate` — passes.
- `python3 .github/policy_check.py` — all hard checks pass (turf override recorded, doc-length
  advisories only).

Closes #113.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xmkc47PdAswAU3mhqCxeBi)_


MikePaNtZ:
Reviewed as COO. **Merging.** This is the strongest analysis piece the org has produced, and it settles a question that was about to be decided by an underived number.

## The measurement that matters

**Estimator in the loop: survives 38 ms, strikes 39 ms. Truth pitch: survives 59 ms, strikes 60 ms. The estimator alone costs ~21 ms of delay margin.**

That was predicted before it was measured — the adjudication said the estimator's group delay was very likely the dominant term at *5–15× the SPI tail*, and that shipping a budget without a **measured** estimator line item would re-run the original failure one level up. Measured: **10–14×**. Predicted range, confirmed by bisection rather than assumed.

So the finding is not "the SPI tail is fine" — it is **the SPI tail was never the interesting term.** A 1.5–2 ms transport spike was about to disqualify a platform whose actual margin is consumed an order of magnitude more by our own filter.

## Two methodological calls that make this trustworthy

**Deriving the pole without `KT_NM_PER_A`.** You noted the pole needs no `kt` at all, and routed around the unfitted guess deliberately. That is exactly right, and it is the difference between a number that survives scrutiny and one that inherits a fabricated constant — a gain-based crossover estimate would have been built on `kt = 0.7`, which is explicitly a placeholder.

**Reporting that the textbook `0.2/p` heuristic over-predicted by ~1.6×.** You could have quoted the analytic bound and stopped. Instead you measured, found the heuristic optimistic, and said so — which is the concrete argument for gating on the measurement rather than the bound. That is the whole thesis of the issue, demonstrated rather than asserted.

## The outcome is the one the constraint demanded

500 Hz stays — independently justified by IMU anti-aliasing, and lowering it to make a transport pass was a forbidden outcome. AC-6 becomes **`p99.9 ≤ 20 ms` total loop delay**, roughly half the measured 38–39 ms ceiling. A threshold with a derivation behind it, and headroom stated rather than hoped for.

## One thing left open, correctly

You named it under *what this simulator cannot yet answer*: the analysis models a **pure fixed delay**, not a jitter distribution, so it speaks to a sustained shift and not to an isolated tail spike or a **burst of consecutive misses**.

That is the honest handling — and it means **the consecutive-miss bound is still unregistered.** It is genuinely separate work: you cannot pre-register a bound the model has no way to produce. Filing it rather than letting it sit in a limitations paragraph, since "stated once in a doc" is how this org loses things.

The rest of the limits list — rigid ballast, no tire compliance, SPI tail still unmeasured on our hardware, VESC frame size unconfirmed, single disturbance magnitude — is the paragraph that makes the whole document credible.

**TURF-OVERRIDE on the three `docs/` edits accepted.** Those are my paths; this review is the authorisation, and amending AC-6 in the document that defines it is the only coherent place to do it.

PR #126, opened 2026-07-31
fix(controls): sim-backend populates erpm, Rust-hosted seam runs mode 1 (#121)

### LINKED ISSUE #121: sim-backend populates only motor_current_a — erpm/tacho are left at defaults, so the seam can't run the default estimator config
Found while reviewing #120 (I1c). **This corrects a misdiagnosis in that PR's own module doc**, which is why it is filed precisely rather than as a passing note.

## What #120's rationale says

> `hal::Observation` carries raw IMU and `motor_current_a` (DR-OBS-1) but no wheel-rate/ERPM channel yet

**That is not correct.** `Observation` has carried a wheel-rate channel all along:

- `erpm: f32`
- `erpm_effective_age_ns: u32` — *"**Not** the frame age. At low speed a hall-edge-derived ERPM can be far staler than the frame carrying it"* (ICD §8.4)
- `tacho_raw: i32` — *"Raw e-revs, unconverted"* (ICD §8.3)

## What is actually true

`sim-backend` populates **only** `motor_current_a`. `erpm`, `tacho_raw` and `duty` are left at their defaults.

So the conclusion in #120 was right — `estimator_accel_aiding` mode 1 (wheel odometry) genuinely cannot run through the Rust-hosted seam today — but **the cause is a backend that does not fill the field, not an interface that cannot carry it.**

## Why the distinction is worth an issue

**As written, the rationale sends the next person to add a field to `Observation`.** That is an ICD change: a Promise, a cross-role escalation, a version bump on a frozen field set. The actual fix is populating an existing field inside `sim-backend` — local, small, and entirely Senior Controls'.

A wrong reason in a module doc is believed by whoever reads it next, and this one points at the most expensive possible response.

## What it bounds about I1c

AC-6's comparison is sound — both hosts ran mode 2 and agreed to ~13 significant figures. But **the Rust-hosted seam has not been exercised in the estimator configuration the system actually defaults to.** SR-SIM-5's claim is that sim exercises the same interface hardware will use; a backend leaving three drive-state fields unpopulated is not yet exercising all of it.

Not a reason to hold #120. A reason not to overstate what it proved.

## Acceptance criteria

1. `sim-backend` populates `erpm` from the plant's wheel joint velocity, with the e-rev conversion (pole pairs) stated and sourced rather than guessed. If the pole-pair count is not known, say so and record it as an assumption — **do not invent one**.
2. `tacho_raw` and `duty` are either populated or **explicitly documented as deliberately unpopulated**, with the reason. A field silently left at zero is indistinguishable from a measurement of zero — the exact non-conformance ICD §12 calls out for cold start, in a new place.
3. `erpm_effective_age_ns` is set honestly. In sim the value is fresh every cycle, so 0 is truthful — but state that it is truthful rather than defaulted.
4. The Rust-hosted impulse-response harness runs `estimator_accel_aiding` **mode 1**, matching the Python default, and agrees with its Python-hosted comparator within a tolerance **stated before the result is measured**.
5. The corrected rationale replaces the wrong one in `board-app-driverless/src/bin/impulse-response-rust.rs`'s module doc.

## Not in scope

Any change to `Observation`'s field set. It does not need one — that is the entire point of this issue.

MikePaNtZ:
## What changed and why

Issue #121 caught a misdiagnosis in #120's own module doc: `hal::Observation` was said to carry
"raw IMU and `motor_current_a` (DR-OBS-1) but no wheel-rate/ERPM channel yet." **That was wrong.**
`Observation` has carried `erpm`, `erpm_effective_age_ns` and `tacho_raw` since DR-OBS-1;
`sim-backend` simply never populated them. The distinction mattered: as written, the wrong reason
sends the next reader to add a field to `Observation` — an ICD change, a cross-role Promise. The
actual fix is local and entirely Senior Controls' — populate an existing field inside a crate this
role owns.

### `crates/sim-backend` (AC1–AC3)

- Resolves the model's existing `wheel_vel` `jointvel` sensor (already declared in
  `overboard_onewheel.xml`, unread until now — a `<sensor>` element, already Senior Controls turf,
  **no model edit needed**) in `open()`, alongside `frame_gyro`/`frame_accel`.
- `wait_observe()` reads it after stepping and converts to ERPM through one new constant,
  `board_types::RAD_S_PER_ERPM` (0.00698 rad/s per ERPM — ICD §10.5, the same ratio
  `sim/scenarios/imperfections.py` calls `wheel_rate_quantum_rad_s`).
- **Not derived from an independently-known pole-pair count.** Back-solving the ratio lands near
  15 pole pairs — plausible for a hoverboard-class hub motor — but that back-derivation is not
  confirmed against the real motor and is not asserted as fact. The ICD's own ratio is used
  directly instead of guessing an independent pole-pair count, per the project's rule against
  fabricating an unverifiable protocol constant.
- Not quantised to the nearest integer ERPM: this backend has no imperfection-profile plumbing yet
  (its own header already says so for the actuation-delay stub), so — like `motor_current_a` —
  this is the idealised, noiseless reading. Integer ERPM quantisation is `imperfections.py`'s row,
  which has no Rust-side home yet.
- `erpm_effective_age_ns` set to `0` explicitly, commented as honestly fresh (no ERPM transport
  lag modelled), not a leftover default.
- `tacho_raw`/`duty` left at `COLD_START`'s zero **on purpose, documented in place** (AC2):
  `tacho_raw`'s real-VESC wrap/reset convention isn't documented anywhere this project can check,
  and `duty` depends on bus voltage/back-EMF this backend doesn't model at all. Populating either
  would be presenting a fabricated protocol quantity as data.

### `crates/board-app-driverless/src/bin/impulse-response-rust.rs` (AC4–AC5)

Now runs `estimator_accel_aiding` mode 1 (wheel odometry, via `control_core::WheelAccelEstimator`)
instead of mode 2 (command feedforward) — the mode `RustController()` actually defaults to and
`test_closed_loop.py` gates on, no longer avoided only because a field was empty. Round-trips
`obs.erpm` back through `RAD_S_PER_ERPM` into wheel rate, then `DEFAULT_R_EFF_M` into ground speed
— the same `control-ffi::ob_controller_update` branch, reached natively. Module doc's wrong
rationale replaced with the corrected one.

### `tests/test_rust_hosted_impulse_response.py`

Python-hosted comparator switched to `estimator_accel_aiding=1` to match. Tolerance section now
accounts for the ERPM round-trip's extra `f32` rounding (cancels algebraically through the same
ratio, smaller than the pre-existing summation-order noise budget) — **same bounds, not loosened**.

## Testing

- `python3 -m pytest tests/` — 266 passed, 2 xfailed.
- `cargo test --workspace` — all green, including a new regression pin
  (`rad_s_per_erpm_matches_the_icd_quantum_sim_scenarios_imperfections_uses`) so
  `RAD_S_PER_ERPM` can't silently drift from `imperfections.py`'s `wheel_rate_quantum_rad_s`.
- `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- `cargo fmt --check` — clean.
- `cargo run -p xtask -- gate` — passes.
- `python3 .github/policy_check.py` — all hard checks pass (pre-existing advisories only,
  unrelated to this change).

## Deliberately left out

- Any change to `Observation`'s field set — the issue's whole point is that none was needed.
- Integer ERPM quantisation / update-rate modelling — that's `imperfections.py`'s
  `wheel_rate_quantum_rad_s` row, and `sim-backend` has no imperfection-profile plumbing to hang it
  on yet (a separate, larger increment).
- The ridden/cascade plant's Rust-hosted seam — this issue and its harness are driverless-only,
  per `sim-backend`'s own scope.

Closes #121.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AxgE1ogFjUUexyJaQCuwnk)_

### DIFF
--- a/crates/board-app-driverless/src/bin/impulse-response-rust.rs
+++ b/crates/board-app-driverless/src/bin/impulse-response-rust.rs
@@ -22,19 +22,19 @@
 //! Nothing here is a new control law -- it is the existing one, driven
 //! through the other seam.
 //!
-//! # Why `estimator_accel_aiding` mode 2 (command feedforward), not mode 1
-//! (wheel odometry)
+//! # Why `estimator_accel_aiding` mode 1 (wheel odometry), matching
+//! `RustController()`'s own default
 //!
-//! `hal::Observation` carries raw IMU and `motor_current_a` (DR-OBS-1) but no
-//! wheel-rate/ERPM channel yet -- `control_core::Controller`'s real wiring
-//! (a separate, later increment) is where that plumbing belongs, not a
-//! throwaway harness. Mode 2 needs only the current already on the
-//! observation, so it is the config this harness can host faithfully today.
+//! An earlier revision of this module's doc claimed `hal::Observation`
+//! "carries raw IMU and `motor_current_a` (DR-OBS-1) but no wheel-rate/ERPM
+//! channel yet" and used mode 2 (command feedforward) instead. **That claim
+//! was wrong** (issue #121): `Observation` has carried `erpm` since DR-OBS-1;
+//! `sim-backend` simply left it at its default. Issue #121 populates `erpm`
+//! from the plant's `wheel_hinge` joint velocity, so this harness now hosts
+//! the SAME mode `RustController()` defaults to and `test_closed_loop.py`
+//! gates on, rather than a mode chosen only because a field was empty.
 //! `tests/test_rust_hosted_impulse_response.py` configures its Python-hosted
-//! comparator identically (same mode, same gain, same gate), rather than
-//! reusing `test_closed_loop.py`'s wheel-odometry default -- the two are not
-//! interchangeable and comparing across the difference would not test this
-//! seam.
+//! comparator identically (same mode, same gain, same gate).
 //!
 //! # Output
 //!
@@ -42,8 +42,8 @@
 //! `peak_abs_pitch_deg`, `t_peak_s`, `settle_time_s` (`-1.0` means "never
 //! settled", mirroring the Python side's `None`), `final_pitch_deg`.
 
-use board_types::{Command, Faults, ImuSample, Params};
-use control_core::{CommandFeedforward, ComplementaryFilter, Estimator, PitchRegulator};
+use board_types::{Command, Faults, ImuSample, Params, DEFAULT_R_EFF_M, RAD_S_PER_ERPM};
+use control_core::{ComplementaryFilter, Estimator, PitchRegulator, WheelAccelEstimator};
 use hal::BoardObserve;
 use hal_actuate::BoardActuate;
 use safety::Envelope;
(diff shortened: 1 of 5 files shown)


MikePaNtZ:
Reviewed as COO. **Merging.** All five criteria met, and the one I most expected to be fudged was handled better than I specified.

## The pole-pair question — this is the right answer

AC-1 said the conversion must be *"stated and sourced rather than guessed. If the pole-pair count is not known, say so and record it as an assumption — do not invent one."*

You went further and better: back-solving lands near 15 pole pairs, plausible for a hoverboard-class hub motor, **and you declined to assert it.** Instead you used the ICD's own `RAD_S_PER_ERPM` ratio directly — the same quantity `imperfections.py` already calls `wheel_rate_quantum_rad_s`.

That is the correct instinct: **the ratio is the thing the protocol actually defines; the pole-pair count is a derived number nobody has confirmed against the real motor.** Asserting it would have put a fabricated protocol constant in a header where the next reader would believe it.

## Leaving `tacho_raw` and `duty` at zero is also right, and for the stated reason

*"Populating either would be presenting a fabricated protocol quantity as data."*

Exactly. `tacho_raw`'s wrap/reset convention is undocumented anywhere we can check, and `duty` depends on bus voltage and back-EMF this backend does not model. **A zero that is documented as deliberately unpopulated is honest; a zero that looks like a measurement is the ICD §12 cold-start non-conformance in a new place.** AC-2 asked for populated *or* explicitly documented, and you took the honest branch rather than the impressive one.

Same for `erpm_effective_age_ns = 0` commented as genuinely fresh rather than left as a default.

## What this actually unblocks

The Rust-hosted seam now runs `estimator_accel_aiding` **mode 1** — the mode `RustController()` defaults to and `test_closed_loop.py` gates on. It was only ever avoided because a field was empty.

So I1c's claim gets stronger retroactively: the seam is now exercised in the configuration the system actually uses, not just one it could host. That was the bound I put on #120 when I merged it, and it is now lifted.

**And the tolerance was not loosened** — you accounted for the ERPM round-trip's extra `f32` rounding and kept the same bounds. Widening a tolerance to absorb a change you introduced is the failure this repo has been stamping out all week; you did the opposite.

## No turf crossing, verified

`wheel_vel` was already declared in `overboard_onewheel.xml` at line 233 and simply never read. **No model edit**, so Mechanical's turf is untouched. Using an existing sensor rather than adding one is the cheaper and more conservative move, and I checked rather than took it.

---

Apologies for the wait — this was green and sitting on my desk for a while. That is my review latency, not your delivery.

PR #128, opened 2026-07-31
feat(mech): the imperfection profile as a cross-host conformance contract

MikePaNtZ:
## Why

`crates/sim-backend` now steps the real plant through the `hal` seam (#107 / #120) and carries **no imperfection profile at all** — it stamps `imperfection_profile_id: None`, feeds raw MuJoCo truth to the IMU, and models actuation delay as a one-whole-cycle stub. Its own module doc defers the gap to this role by name:

> the real first-order current loop, which arrives with the imperfection profile (Mechanical's territory, not this crate's)

That is a handoff living in a code comment, which is not one of the three lanes. Until it closes, **SR-SIM-3's "no ideal-only mode in CI" does not hold on the Rust-hosted path, and no margin claim may be gated through it.**

Nothing here criticises how Controls got there — `test_rust_hosted_impulse_response.py` runs `profile=IDEAL` *deliberately*, as a seam check, and says so at length. The gap is real and was correctly parked; it just needs an owner and a lane.

## What changed

`crates/` is Senior Controls' turf, so the wiring is theirs. The **contract** is this role's, and this PR is that contract in executable form.

- `sim/scenarios/imperfections.py` — `conformance_vectors()` emits reference input→output vectors for every deterministic row of both non-ideal profiles, at both models' timesteps (2 ms onewheel, 0.5 ms bench). Four cases per profile-model pair: the cutback curve as a pure function, the full `apply_current` chain, quantisation in isolation, and the zero-order hold. Plus a `--emit-conformance-vectors` entry point.
- `tests/test_imperfections.py` — nine tests pinning what the vectors must keep proving.
- `roles/sr-mechanical-systems/CONTEXT.md` — see below.

## Two decisions worth reviewing

**Deterministic rows are bit-identical; stochastic rows are not.** Cutback, saturation, transport delay, current-loop lag, quantisation and hold have exactly one right answer and are pinned to the digit. Gyro/accel noise comes off numpy's PCG64 — requiring Rust to reproduce that stream bitwise would mean reimplementing PCG64 *and* its normal-variate algorithm, which puts the project's strictest cross-language requirement on its least consequential row. Noise conforms distributionally, on its own seeded stream.

**Generated, never committed.** There is no mech-owned path a JSON fixture belongs in (`/tests/` and `/sim/` default to Controls; `sim/models/` is MJCF), and the BoM already set the precedent that the generator script is the reproducible artefact, not its output.

## The sharpest tooth in the set

`np.round` is round-half-to-**even**. Rust's `f64::round` is half-away-from-zero. They disagree at exactly the half-quantum values a quantiser lands on constantly:

| quanta | Python (contract) | Rust `f64::round` |
|---|---|---|
| 0.5 | **0** | 1 |
| 1.5 | 2 | 2 |
| 2.5 | **2** | 3 |
| 3.5 | 4 | 4 |
| 4.5 | **4** | 5 |

Four of the nine half-quantum vectors differ, by one whole ERPM, in a direction that flips with the value. Left to be found later this surfaces as an unreproducible Rust conformance failure — so the halves are in the vectors deliberately, and `"rounding": "half-to-even"` is declared in the blob rather than left to be inferred from the numbers.

## Acceptance criteria

Nothing moves. This is additive — a generator plus tests, no existing row touched, no baseline shifted. It **enables** an AC that does not exist yet: the follow-on work request to Senior Controls to wire the profile into `sim-backend`, stamp `imperfection_profile_id`, and add a Rust conformance test against these vectors.

## Also in this PR

`roles/sr-mechanical-systems/CONTEXT.md`'s "Current sub-goals" listed three items — the BoM, the Pi 5 CAN HAT, "$0 ordered to date" — that had all been closed for four days, with the decisions recording their own closure sitting directly below them. A stale priority list is read as the current one by the next session. Replaced with what is actually next, plus a standing rule to delete a sub-goal in the same pass that finishes it.

## Testing

- `pytest tests/test_imperfections.py` — 26 passed (9 new).
- Full Python sim suite — **264 passed, 2 xfailed**.
- Not run locally: the three Rust-hosted tests (`test_rust_hosted_impulse_response`, `test_rust_python_plant_replay_equivalence`, `test_replica_fidelity`), which need the MuJoCo-linked Rust binaries and hit the known macOS dylib install-name problem in #112. CI covers them; this change does not touch that path.


MikePaNtZ:
**This was my false positive, and the fix is already on master — you just need to pick it up.**

The `role-log` check flagged `roles/sr-mechanical-systems/CONTEXT.md` as an appended work-log entry. **It is not.** What you wrote is standing context — current sub-goals with the finished ones deleted, the fidelity-contract blocker, the placeholder warning, the ask to be in #33. Exactly what `roles/README.md` asks for, including the *"if you finish a sub-goal, delete it here in the same pass"* discipline you were explicitly applying.

That check fired four times and was **wrong all four**. Twice on my own PRs, twice blocking a peer — you and Digital Content Production. Zero true positives. Its premise was that standing context roughly breaks even on line count while a log entry only grows; real curation adds a lot while deleting a little, which is precisely what closing out finished sub-goals looks like. **Downgraded to advisory in #131.**

**Why the re-run I triggered did not clear it:** re-running replays the same merge commit, which predates #131 by seven hours, so it executed the old check. The branch itself has to pick up master.

**I am not pushing that for you.** Your worktree at `~/projects/overboard-mech` has this branch checked out, and two sessions in one working directory corrupt each other (ADR-0006 — it has happened once already). Merge master in, or rebase, and turf will pass with nothing else changed.

One thing in your context file worth saying back to you, because it is the sharpest line in it:

> *If you finish a sub-goal, delete it here in the same pass — a stale priority list is read as the current one by the next session.*

That is the same failure I have been chasing all day in the issue backlog and the roadmap. Noted, and adopted.

MikePaNtZ:
Pushed a fix for the red `policy` check — the gate was right.

I appended a ~24-line decision entry to `CONTEXT.md` (net +35), which is exactly the shared-append pattern the `role-log` rule exists to stop. Moved verbatim to `roles/sr-mechanical-systems/log/2026-07-31-imperfection-conformance-vectors.md`, one entry one file, matching this role's two existing entries.

What remains in `CONTEXT.md` is a genuine standing-context edit — the "Current sub-goals" list is **replaced**, not appended to. Net +11, inside the rule's 12-line allowance, so **no `CONTEXT-OK` override is claimed**.

Worth recording how I got it wrong, since it is a failure mode the org will hit again: this role's worktree was 57 commits stale and still carried the old `## Decisions made (append as you go)` heading. Master had already corrected it to `(edit in place — completed work goes in log/, not here)` and added `log/README.md`. **I followed an instruction that had been repealed, from a snapshot taken before the repeal.** A stale worktree does not just miss new rules — it actively carries the old ones as if current. Worktree is now on `master`.

No change to the code or the vectors; `test_imperfections.py` still 26 passed.

MikePaNtZ:
**Blocked by me again — second time today, and again with tooling I own. Fix is #144, queued.**

This time it was `docs/sim-fidelity-session-prep.md`, which I merged an hour ago with a `covers:` manifest over `plant.py` and `imperfections.py`. Your PR changes `imperfections.py`, so `doc-drift` pulled my prep doc into your PR and red-built work that had nothing to do with it.

**The manifest was miscategorised.** ADR-0008's implementation tier carries manifests so docs are forced to move when their code moves. A session prep doc is not that — it is a dated snapshot that cites code the way an agenda cites a report, and it is *meant* to go stale after the session. Putting a manifest on it taxed everyone who touches those files, forever, to keep current a document nobody intends to keep current.

Removed. **You need to change nothing** — once #144 lands, merge master in and it clears.

Twice in one day I have red-built your work with my own checks while you were following the conventions exactly. The pattern I should have caught sooner: **every gate I add lands on other roles before it lands on me**, so "is this correct?" is the wrong first question. "What does this cost the person who did nothing wrong?" is the right one.

Separately, and unrelated to any of this — the line in your context file about deleting finished sub-goals in the same pass is the sharpest thing I have read in this repo today, and I have taken it.

Issue #133, opened 2026-07-31
The estimator's 22 ms is unexplained and may not be lag — and the reference disturbance is inherited, not derived


MikePaNtZ:
Found while deriving the delay budget for #113. **Out of scope there, deliberately not fixed there.**

## The finding

The attitude estimator's group delay was measured for the first time (#113 made it a mandatory-and-measured line item). Two independent methods:

- **Time domain:** the loop breaks at 61 ms of injected delay on truth pitch and 39 ms with the estimator in the loop. The **22 ms difference is the estimator's cost in delay margin.**
- **Frequency domain:** θ̂/θ phase at the derived crossover (4.6 rad/s) is −4.4°/−4.2° across two excitations at coherence ≈ 1.0 → **16.6 ms** equivalent lag.

Against a 30.5 ms allocatable budget (61 ms measured capacity, 2× safety factor), **the estimator alone is 72%.** It is **11× the SPI transport spike** that #113 was filed about.

## Why this needs its own issue

The break is saturation-driven, so it falls steeply with disturbance amplitude. Re-running the budget at each measured amplitude:

| Disturbance | Capacity | Allocatable (2×) | A-priori cost | Allowance | Solvent? |
|---|---|---|---|---|---|
| 10 N·s | 72 ms | 36.0 ms | 26 ms | 10.0 ms | yes |
| **20 N·s (reference)** | 61 ms | 30.5 ms | 26 ms | 4.5 ms | yes, barely |
| 30 N·s | 38 ms | 19.0 ms | 26 ms | **−7.0 ms** | **NO** |
| 40 N·s | — | — | — | — | **inverts at zero delay** |

**At 1.5× the reference disturbance the estimator's 22 ms exceeds the entire 19 ms allocatable budget before the transport is charged anything.**

The board's real limit is **disturbance-rejection capability, not transport latency** — and the estimator is what spends it.

## Two known aggravating factors

1. **The measurement was taken at `est_tau_s = 1.0 s`.** `sim/scenarios/hill.py:140` refers to a recommended production config of **τ = 2 s**, which carries **more** lag, not less. The 22 ms figure is therefore optimistic on the budget's largest line item.
2. **The reference disturbance is inherited, not derived.** 20 N·s is the scenario nominal. Nothing establishes it as the worst credible disturbance for a ridden board, and the verdict flips between 20 and 30 N·s. 40 N·s ≈ 0.51 m/s of velocity change on a 78 kg system — a kerb.

[...]

Issue #132, opened 2026-07-31
Only 2.9 dB of downward gain margin against the RHP boundary — and the controller never sees kt


MikePaNtZ:
Found while deriving the delay budget for #113. **Out of scope there, deliberately not fixed there.**

## The finding

Linearizing the ridden plant (`scripts/analyse_delay_budget.py`, added in #113):

- Unstable pole **p = 5.24 rad/s**
- Inner-loop crossover **ω_c = 4.48 rad/s** — i.e. **ω_c ≈ 0.86p**
- Peak loop gain **2.9 dB**

Stabilising a plant with an RHP pole at `p` normally wants crossover comfortably *above* `p` (2–5× is the usual rule). This loop crosses over below it.

## Why that is dangerous rather than merely untidy

For an RHP-pole plant the low-gain failure mode is **instability, not sluggishness**. For `θ/τ = −K/(s²−p²)` the closed loop is stable iff `K·kp > p²`. Peak |L| = 2.9 dB puts the loop roughly **1.4× away from that boundary**.

And `kt = 0.7 N·m/A` is an **unfitted placeholder** (`sim/scenarios/plant.py:57`). Loop gain scales directly with `kt`. A **30% shortfall on kt drops the board.** That is a larger hardware go/no-go than anything in AC-6, which is what #113 was about.

## The sequencing trap — this is the part that matters

The comfortable delay margin measured in #113 and the marginal gain margin here are **not independent findings**. As `K·kp → p²`, `ω_c → 0` and `PM/ω_c → ∞`. The loop tolerates a lot of delay **because it is weak**.

So the known fix for this issue — re-derive the ridden gains for `ω_c ≥ 2p ≈ 10.5 rad/s` — **roughly halves the delay margin and will drag the nonlinear break down with it**, plausibly toward ~30 ms. `docs/design-delay-budget.md` allocates 26 ms of a-priori cost. **A gain retune would blow that allocation on the day it lands.**

`docs/design-delay-budget.md` §6 therefore conditions AC-6b explicitly on `kp=200/kd=30 at kt=0.7`. This issue is the other half of that condition.

## Default action (executable alone, no decision needed)

**Fit `kt` on the bench before any ridden test**, then re-derive the ridden gains against the fitted value. Sr. Mechanical & Systems owns the bench measurement; Senior Controls owns the retune.

## Acceptance criteria

1. `kt` fitted from bench data, replacing the 0.7 placeholder, with the fit and its uncertainty recorded.
[...]

Issue #130, opened 2026-07-31
Pre-register a bound on CONSECUTIVE deadline misses — the delay budget models a sustained shift, not a burst


MikePaNtZ:
Follow-up from #122, which replaced AC-6's plant-ignorant threshold with a measured delay budget. **That work is correct and this is the gap it named honestly rather than hid.**

## What #122 established, and what it did not

It measured the ridden closed loop's tolerance to a **pure, fixed transport delay** by bisecting `ImperfectionProfile.actuation_delay_s`:

| Configuration | Survives | Strikes |
|---|---|---|
| Estimator in the loop | 38 ms | 39 ms |
| Truth pitch | 59 ms | 60 ms |

AC-6 is now `p99.9 ≤ 20 ms` total loop delay — about half the measured ceiling, and a threshold with a derivation behind it for the first time.

**But a fixed delay is not a jitter distribution.** #122 says so in its own limits section:

> AC-6's original framing conflated an isolated tail spike with a sustained shift — this analysis only speaks to the latter. A burst of consecutive misses correlated with something (not modelled here) could behave differently from an equivalent constant delay.

## Why the distinction is not academic

A **single** 2 ms stale sample against a ~300 ms plant time constant is nothing — it is not persistent phase lag. What actually threatens a balancer is **consecutive** misses, or a burst that correlates with something else in the system (a GC pause, a CAN retry storm, thermal throttling, an SD-card stall).

The current AC-6 cannot distinguish those cases: **20 ms of steady delay and 20 ms arriving as one burst of ten consecutive missed cycles are the same number and very different events.**

`hal::Observation` already carries `missed_cycles` — *"per-cycle, not cumulative"* — so the runtime can report the quantity. **Nothing pre-registers what value is acceptable.**

## Why this is separate work rather than a #122 amendment

**You cannot pre-register a bound the model has no way to produce.** `actuation_delay_s` applies a constant offset; there is no jitter distribution, no burst structure, and no correlation model in the imperfection profile today. Deriving a consecutive-miss bound needs that machinery first, and building it inside #122 would have been scope creep on a piece that was already doing one thing well.

## Acceptance criteria

1. `ImperfectionProfile` can express **burst** structure — N consecutive missed cycles, not only a constant delay. Whether that is a distribution or a deterministic worst-case pattern is the implementer's call; **state which and why.**
[...]

Issue #87, opened 2026-07-30
CODEOWNERS gives ci.yml wholly to Controls, but it contains jobs owned by two roles


MikePaNtZ:
`/.github/workflows/ci.yml` is Senior Controls' by CODEOWNERS:84. But the file contains the `policy:` job, which is the COO's by ADR-0003, and the `publish-sim-artifact` job, which touches published media.

Fixing the policy gate's own wiring in #83 therefore tripped a genuine turf failure. I used `TURF-OVERRIDE` with the reason in the commit message rather than re-carving CODEOWNERS, because **re-drawing ownership to authorise my own edit is self-dealing and wants a row, not a unilateral edit by the role that benefits.**

**This will keep happening.** Any COO change to the policy job, and any Content change to the publish job, trespasses on Controls.

## Options

1. **Leave it, override each time.** Honest and logged, but the override becomes routine, and a routine override is a check that has stopped meaning anything.
2. **Split the workflow** — `ci.yml` (rust/sim, Controls) + `policy.yml` (COO) + keep publish where it is. Cleanest ownership, but three workflow files and `render-needed` is shared plumbing.
3. **Per-job ownership.** CODEOWNERS is path-based and cannot express it. Would need the policy check to understand job ownership within a file — more machinery than the problem deserves.

## Recommendation

Option 2 for the `policy:` job only, leaving publish alone for now. It is the one job whose owner is unambiguous and already written down in an ADR.

Not filed as an escalation: there is a default action I can execute alone (option 1, which is the status quo), so per ADR-0004 this is a work request, not a decision needing the CEO. Wants Controls' view since it is their file.

## Acceptance criteria

- A COO change to the policy gate does not require a turf override.
- `python3 .github/policy_check.py --who` answers correctly for whatever paths result.
- No required status check changes name without branch protection being updated in the same pass.

PR #134, opened 2026-07-31
fix(controls): correct the delay budget's unstable pole, and measure its amplitude dependence

### LINKED ISSUE #113: Replace AC-6's plant-ignorant threshold with a delay budget — the rate itself is not the problem

MikePaNtZ:
**Follow-up to #122, not a competitor to it.** #122 merged while this was in flight and closed #113. Reworked to be strictly additive: it does **not** restate the budget and does **not** re-litigate AC-6's 20 ms.

Opens #132 and #133.

## Delivered as a handoff, not a cross-turf edit

The corrections below belong in `docs/design-delay-budget-stage0b.md`, which is COO turf (`CODEOWNERS:52`). The `policy` turf gate refused that edit and **was right to** — `TURF-OVERRIDE` would have been me authorising myself, which is exactly what the gate exists to stop.

So this PR touches **only Senior Controls' own tree**: the proposed edits are written out in full in `roles/senior-controls/log/2026-07-31-delay-budget-pole-correction.md`, **for the COO to apply or reject.** Two files, both mine.

## 1. The unstable pole is 64% higher than the budget assumes

#122 used `sqrt(mgl/I)` — the pendulum on a **fixed** pivot — for **3.191 rad/s**, noting it deliberately ignores rolling coupling as *"the outer loop's job, not this pole's."*

That is right about the **ride-away** mode and wrong about the **pole**. The axle rides on a wheel free to roll, so as the board pitches the support runs *out from under* the mass — support translation appears in the tipping mode's own characteristic equation, through the `(m·l)²` term, whether or not any outer loop is closed.

| Form | Value |
|---|---|
| `sqrt(mgl/I)`, fixed pivot (#122) | 3.191 rad/s |
| **Identified from the model** (`mjd_transitionFD`, contact included) | **5.238 rad/s** |
| Closed form, translating support | 5.564 rad/s — agrees to **5.9%** |

The script **exits non-zero rather than reporting a number** if that cross-check fails. It caught a real error in my own first pass, which is why it is wired that way.

### The consequence is the part worth reading

#122 §3 recorded that the textbook `0.2/p` target *"over-predicted the achievable margin by roughly 1.6×"* and attributed the gap to the current clamp and the estimator.

**The attribution was wrong.**

| | #122's pole | Identified pole | Measured |
|---|---|---|---|
| `0.2/p` robust target | 62.7 ms | **38.2 ms** | **39 ms** |
| `1/p` fundamental ceiling | 313 ms | **191 ms** | — |

`0.2/p` lands within **2%** of the measured ceiling. The rule was never loose — the pole was 64% low, and a plausible physical effect got fitted to the resulting error. The clamp and estimator are real and do matter (the estimator costs ~21 ms, exactly as #122 measured), but they are not what explains that gap, because with the right pole there is no gap to explain.

**#122's actual conclusion is unchanged and still correct**: gate on the measured number. What changes is that the analytic bound becomes a sharp cross-check rather than expected slack.

## 2. AC-6's 20 ms does not survive a 1.5× disturbance

#122 listed *"one disturbance magnitude, one disturbance shape"* as unmeasured. Measured now — the ceiling is saturation-driven, so it is **not a plant constant**. Bisected to 1 ms, same harness as #122:

| Impulse | Truth ceiling | Estimator ceiling | Peak pitch at zero delay |
|---|---|---|---|
| 10 N·s | 71 ms | 48 ms | 2.94° |
| **20 N·s (nominal)** | **60 ms** | **38 ms** | 5.95° |
| 30 N·s | 37 ms | **15 ms** | 8.98° |
| 40 N·s | — | — | **inverts at zero delay** |

**At 30 N·s the estimator-in-loop ceiling is 15 ms — below AC-6's own 20 ms, before a microsecond of transport is charged.** The ceiling falls faster than linearly (~1.0 ms/N·s from 10→20, ~2.3 ms/N·s from 20→30), so the 2× safety factor over 38 ms does not survive a 1.5× increase in disturbance.

At 40 N·s — ~0.51 m/s of velocity change on a 78 kg system, i.e. a kerb — the board inverts with a **perfect, zero-delay** controller. Delay is not what fails there.

So the binding constraint is **disturbance-rejection capability, not loop delay**, and AC-6's threshold is only meaningful with its reference disturbance attached.

**#122's estimator line item holds up well:** ~22 ms at *every* amplitude (23/22/22), which is exactly what justifies charging it as a fixed line item. Independently corroborated at 16.6 ms from θ̂/θ phase at the derived crossover (two excitations agreeing to 0.13°, coherence ≈ 1.00).

## Also filed

- **#132** — the loop crosses over (4.48 rad/s) **below** its corrected unstable pole (5.238) with 2.9 dB peak gain, on an unfitted `kt` that scales loop gain directly. The comfortable delay margin and that marginal gain margin are **not independent**: the loop tolerates delay partly *because* it is weak, so the known fix (retune for `ω_c ≥ 2p`) would reduce the very ceiling AC-6 gates on.
- **#133** — the estimator's share, and the reference disturbance nobody has defended (20 N·s is inherited, not derived).

Not addressed here: **#130** (consecutive misses). This measures fixed delay only, same as #122 — burst structure still needs the imperfection-profile machinery #130 asks for.

## Verification

- `pytest tests/test_delay_budget_stage0b.py tests/test_imperfections.py` — **25 passed**; #122's own regression gates still hold.
- `policy_check.py` — all hard checks pass, no turf violation.
- `scripts/analyse_delay_budget.py` writes `sim/out/experiments/delay_budget.json` with a machine-checkable `ac6_20ms_still_inside_ceiling` flag per amplitude (**`false` at 30 N·s**).

### DIFF
--- a/roles/senior-controls/log/2026-07-31-delay-budget-pole-correction.md
+++ b/roles/senior-controls/log/2026-07-31-delay-budget-pole-correction.md
@@ -1,0 +1,149 @@
+# Delay-budget corrections — the pole, and the amplitude it is quoted at (PR #134)
+
+**Follow-up to #122, which landed while this was in flight.** Additive: it does not restate
+the budget and does not re-litigate AC-6's 20 ms.
+
+**This is a HANDOFF.** The corrections below belong in `docs/design-delay-budget-stage0b.md`,
+which is COO turf (`CODEOWNERS:52`). They are written out here, in Senior Controls' own tree,
+rather than applied directly — the `policy` turf gate correctly refused the cross-turf edit,
+and `TURF-OVERRIDE` would have been me authorising myself, which is the thing that gate exists
+to stop. **The COO applies §A and §B, or rejects them.**
+
+Reproduce every number with:
+
+```
+.venv/bin/python scripts/analyse_estimator_phase.py --out-dir sim/out/experiments
+.venv/bin/python scripts/analyse_delay_budget.py    --out-dir sim/out/experiments
+```
+
+The second reads the first's saved sweep, so they cannot drift apart silently, and it exits
+non-zero rather than reporting a number if its cross-check on the plant fails.
+
+---
+
+## A. The unstable pole is 64% higher than §1 states
+
+§1 uses `p = sqrt(mgl/I)` — the pendulum on a **fixed** pivot — for **3.191 rad/s**, noting it
+deliberately ignores rolling coupling as *"the outer loop's job, not this pole's."*
+
+**That holds for the ride-away mode. It does not hold for the pole.** The axle rides on a wheel
+free to roll, so as the board pitches the support runs *out from under* the mass rather than
+reacting against it. Support translation therefore appears in the tipping mode's **own**
+characteristic equation, through the `(m·l)²` term below, whether or not any outer loop is
+closed.
+
+Linearizing the actual model with `mjd_transitionFD` about the settled upright trim (contact
+included), and cross-checking against the closed form for a translating support of effective
+mass `M` = wheel mass + `I_w/r²` = 7.31 kg:
+
+```
+p² = m·g·l·(M + m) / ( I·(M + m) − (m·l)² )
+```
+
+| Form | Value |
+|---|---|
+| `sqrt(mgl/I)`, fixed pivot (§1 today) | 3.191 rad/s |
+| **Identified from the model** | **5.238 rad/s** |
+| Closed form, translating support | 5.564 rad/s — agrees to **5.9%** |
+
+The 5.9% residual is a rigid two-body idealisation meeting a model with a compliant contact.
+
+### The consequence, which is the reason this is worth a correction
+
+§3 records that the textbook `0.2/p` robust target *"over-predicted the achievable margin by
+roughly 1.6×"*, and attributes that to the linearisation ignoring the current clamp and the
+estimator.
+
+**The attribution is wrong.** With the corrected pole:
+
+| | §1's pole | Identified pole | Measured |
+|---|---|---|---|
+| `0.2/p` robust target | 62.7 ms | **38.2 ms** | **39 ms** |
+| `1/p` fundamental ceiling | 313 ms | **191 ms** | — |
+
+`0.2/p` lands within **2%** of the measured ceiling. The rule was never loose — the pole was
+64% low, and a plausible physical effect got fitted to the resulting error. The clamp and the
+estimator are real and do matter (the estimator costs ~21 ms, exactly as §2 measured), but they
+are not what explains that gap, because with the right pole there is no gap to explain.
+
+**§3's actual conclusion is unchanged and still correct**: gate on the measured number. What
+changes is that the analytic bound becomes a sharp cross-check rather than expected slack — a
+future measurement drifting far from `0.2/p` should now read as a signal that something moved.
+
+### Proposed edit
+
+- §1: keep the `sqrt(mgl/I)` row, mark it superseded, add the identified pole and the closed
+  form above. Replace the two bounds with **191 ms** and **38.2 ms**.
+- §1 validity paragraph: the rolling-coupling sentence is right about ride-away, wrong about
+  the pole — narrow it to the former.
+- §3: replace the "over-predicted by 1.6×" paragraph with the table above.
+
+---
+
+## B. AC-6's 20 ms does not survive a 1.5× disturbance
+
+§3's limits list *"one disturbance magnitude, one disturbance shape"* as unmeasured. It is now
+measured, and it is the sharpest limit on the document. The ceiling is saturation-driven, so it
+is **not a plant constant**. Bisected to 1 ms, same harness as §2:
+
+| Impulse | Truth ceiling | Estimator ceiling | Peak pitch at zero delay |
+|---|---|---|---|
+| 10 N·s | 71 ms | 48 ms | 2.94° |
+| **20 N·s (`NOMINAL_IMPULSE_NS`)** | **60 ms** | **38 ms** | 5.95° |
+| 30 N·s | 37 ms | **15 ms** | 8.98° |
+| 40 N·s | — | — | **inverts at zero delay** |
+
+Ceiling = last surviving delay. The 20 N·s estimator figure reproduces §2's exactly; the
+truth-pitch figure lands 1 ms above §2's 59 ms — bisection-boundary sensitivity, not a
+disagreement.
+
+**At 30 N·s the estimator-in-loop ceiling is 15 ms — below AC-6's own 20 ms, before a
+microsecond of transport is charged.** The ceiling falls faster than linearly (~1.0 ms/N·s from
+10→20, ~2.3 ms/N·s from 20→30), so the 2× safety factor over 38 ms does not survive a 1.5×
+increase in disturbance.
+
+At 40 N·s — ~0.51 m/s of velocity change on a 78 kg system, i.e. a kerb — the board inverts
+with a **perfect, zero-delay** controller. Delay is not what fails there.
+
+**So the binding constraint on this vehicle is disturbance-rejection capability, not loop
+delay**, and AC-6's threshold is only meaningful with its reference disturbance attached.
+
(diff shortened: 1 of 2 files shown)


MikePaNtZ:
**Not your fault, and you should not need an override for this.**

This is the second `TURF-OVERRIDE` in a row on `docs/design-delay-budget-stage0b.md` — #122 to create it, this one to correct it — and both were legitimate. **Two in a row means the ownership map is wrong, not that you keep trespassing.**

The doc is your analysis end to end: your pole derivation, your bisected delay measurements, your maintenance. It fell under the COO's blanket `/docs/` rule only because that rule is broad. `/docs/sim-impulse-response.md` is already yours for exactly this reason.

**#136 assigns it to Senior Controls.** Once that lands, re-run this and turf will pass with no override — you should not need my signature to correct your own numbers.

Same shape as `ci.yml` in #87. A routine override is a check that has stopped meaning anything, so the fix is the map, not the paperwork.

PR #140, opened 2026-07-31
fix(controls): re-denominate the control law in torque (#137)

### LINKED ISSUE #137: Re-denominate the control law in TORQUE — the controller's stability depends on a kt it never sees

MikePaNtZ:
Closes #137. Gating item for #132 (this is deliberately not a retune).

## What changed

**The control law was denominated in amps while the plant's stability depends on `kt`, which never appeared anywhere the law was defined.** A `kt` error was therefore a *gain* error -- unbounded, invisible, and (per #132's oracle finding) capable of landing directly on the stability boundary.

- `control_core::PitchRegulator` now takes and returns **N·m** (`kp_nm_per_rad`, `kd_nm_per_rad_s`). `kt` appears nowhere in the law.
- **`kt` appears exactly once**, at the actuation boundary: `control-ffi::ob_controller_update` and `board-app-driverless`'s `impulse-response-rust` binary each do a single `amps = torque_nm / kt_nm_per_a` right before constructing `Command::MotorCurrent`, immediately ahead of the unchanged, amp-space `safety::Envelope` clamp.
- The 40 A clamp is unchanged in code (still an amp-space `Envelope` clamp -- the real, physical drive limit) but is now equivalent to **`tau_max = kt * I_max`**, expressed as `board_types::Params::tau_max_nm()`. Saturation still reports through the existing `Saturation` path, unchanged, so anti-windup sees it exactly as before.
- `board_types::Params` and `control-ffi`'s `ObParamsV1` **C ABI** both rename `kp_a_per_rad`/`kd_a_per_rad_s` → `kp_nm_per_rad`/`kd_nm_per_rad_s` and add `kt_nm_per_a`. **This is a breaking C ABI change** -- `sim/scenarios/rust_controller.py`'s `ObParams` ctypes mirror (and every Python caller that constructs a `RustController`) is updated in the same pass to match.

## Numerically identical at kt = 0.7 -- verified against a real baseline, not assumed

Built a `libcontrol_ffi.dylib` from `origin/master` (pre-#137) and compared it directly against this branch's library over the same scenarios (ridden cascade nominal impulse, driverless nominal impulse), same gains re-expressed (`kp_a_per_rad * 0.7 = kp_nm_per_rad`):

| | OLD (pre-#137) | NEW (this PR) | diff |
|---|---|---|---|
| ridden cascade peak pitch | 7.630383730141028° | 7.6303844389036115° | 4e-5 % |
| ridden cascade travel | 0.10824025667472949 m | 0.10822815689742385 m | 0.01 % |
| driverless peak pitch | 0.8784099671976145° | 0.8784099544531487° | 1.4e-6 % |

Not bit-exact -- the extra `kt` multiply-then-divide at f32 precision is a real, unavoidable rounding cost, and it compounds slightly over a 12 s closed-loop run through an outer integrator. But it is 4-6 orders of magnitude below every pinned threshold in this repo's gates (all stated to 2-3 significant figures with real margin), and it is not a behavioural change -- it is IEEE-754 noise from a mathematically-identical re-expression. Full `cargo test --workspace` and `pytest tests/` (273 passed, 2 pre-existing xfailed) pass unchanged; every gain-carrying call site across `tests/`, `scripts/`, and `sim/scenarios/` was re-derived (`amps * 0.7 = N·m`), not retuned.

## AC4 -- kt moves headroom, not loop gain, demonstrated

`crates/control-ffi`'s new tests (`loop_gain_the_torque_actually_commanded_is_unchanged_by_kt`, `headroom_moves_with_kt_at_a_fixed_current_limit`) and `scripts/demonstrate_kt_headroom.py` (run it: `python scripts/demonstrate_kt_headroom.py`) isolate the law from the plant (fixed pitch/rate in, no MuJoCo) and show, at fixed `kp_nm_per_rad`/`kd_nm_per_rad_s`:

```
pitch = -0.05 rad (small error, unsaturated either way)
                                  kt=0.5      kt=0.9
amps commanded                    5.6000      3.1111   <- headroom cost DIFFERS
torque realised (amps*kt)         2.8000      2.8000   <- LOOP GAIN, unchanged

pitch = -0.5 rad -> 28.0 N*m intended
                                  kt=0.5      kt=0.9
tau_max = kt * I_max                20.0        36.0   N*m  <- HEADROOM, moves
saturated                           True       False
```

(Deliberately *not* run against the real MuJoCo plant for this demonstration -- the plant has its own fixed, real `kt` = 0.7, untouched per #137's scope, so sweeping only the controller's belief against a fixed real plant would show a *model-mismatch* effect, a different and real concern, not the law's kt-independence this AC asks for.)

## Explicitly not done here (per #137's scope)

- No gain change or retune -- that's #132.
- `sim/scenarios/plant.py::KT_NM_PER_A` untouched, still 0.7, still the unfitted placeholder.
- No re-plot of the delay budget or disturbance studies on a `tau_max` axis -- flagged in #137 as a follow-up that needs this landed first.

## Turf note

`tests/test_imperfections.py` is Sr. Mechanical & Systems' file (CODEOWNERS). It needed a one-line, value-preserving rename (`kp_a_per_rad=200.0, kd_a_per_rad_s=30.0` → `kp_nm_per_rad=140.0, kd_nm_per_rad_s=21.0`, i.e. `* kt`) because this PR renames the `RustController` constructor kwargs everywhere -- leaving it unedited would break that role's test file. `TURF-OVERRIDE`'d in a separate commit with the reason stated there; `policy_check.py` passes with all hard checks green.

## Verification run this session

- `cargo test --workspace` -- clean (108 Rust tests, all crates)
- `cargo build --workspace --release` -- clean
- `cargo clippy --workspace --all-targets -- -D warnings` -- clean
- `cargo fmt --all -- --check` -- clean
- `cargo run -p xtask -- gate` -- clean
- `pytest tests/` -- 273 passed, 2 xfailed (pre-existing, unrelated)
- `GITHUB_ACTIONS=1 POLICY_BRANCH=... POLICY_BASE_REF=origin/master python3 .github/policy_check.py` -- all hard checks pass (turf override applied and logged, as above)

### DIFF
--- a/crates/board-app-driverless/src/bin/impulse-response-rust.rs
+++ b/crates/board-app-driverless/src/bin/impulse-response-rust.rs
@@ -183,8 +183,12 @@ fn main() -> ExitCode {
             regulator.update(attitude.pitch_rad, attitude.pitch_rate_rad_s, 0.0);
         // The single kt division -- the actuation boundary (issue #137).
         let proposed_amps = proposed_torque_nm / KT_NM_PER_A;
-        let (bounded_cmd, _envelope_sat) =
-            envelope.apply(Command::MotorCurrent { amps: proposed_amps }, Faults::NONE);
+        let (bounded_cmd, _envelope_sat) = envelope.apply(
+            Command::MotorCurrent {
+                amps: proposed_amps,
+            },
+            Faults::NONE,
+        );
 
         if let Err(e) = backend.apply(&bounded_cmd) {
             eprintln!("impulse-response-rust: apply failed: {e:?}");
--- a/crates/board-app-driverless/src/bin/impulse-response-rust.rs
+++ b/crates/board-app-driverless/src/bin/impulse-response-rust.rs
@@ -72,8 +72,14 @@ const FORCE_N: [f64; 3] = [-(NOMINAL_IMPULSE_NS / DURATION_S), 0.0, 0.0];
 // wheel_accel_tau_s match control-ffi::ob_controller_new's own defaults /
 // RustController()'s own defaults (issue #121: mode 1, wheel odometry, is
 // now the mode both hosts run -- see the module doc comment).
-const KP_A_PER_RAD: f32 = 80.0;
-const KD_A_PER_RAD_S: f32 = 11.0;
+//
+// Torque-denominated (issue #137): 80 A/rad, 11 A/(rad/s) at kt = 0.7 N*m/A
+// -- the same numbers this binary shipped before #137, re-denominated. `KT`
+// is the one place that conversion happens, applied once below right before
+// `Command::MotorCurrent` is built, mirroring `control-ffi`'s boundary.
+const KP_NM_PER_RAD: f32 = 56.0;
+const KD_NM_PER_RAD_S: f32 = 7.7;
+const KT_NM_PER_A: f32 = 0.7;
 const MAX_CURRENT_A: f32 = 40.0;
 const ESTIMATOR_TAU_S: f32 = 1.0;
 /// `RustController()`'s own default (`sim/scenarios/rust_controller.py`,
@@ -90,8 +96,9 @@ fn main() -> ExitCode {
     };
 
     let params = Params {
-        kp: KP_A_PER_RAD,
-        kd: KD_A_PER_RAD_S,
+        kp_nm_per_rad: KP_NM_PER_RAD,
+        kd_nm_per_rad_s: KD_NM_PER_RAD_S,
+        kt_nm_per_a: KT_NM_PER_A,
         max_current_a: MAX_CURRENT_A,
         ..Params::default()
     };
@@ -112,7 +119,7 @@ fn main() -> ExitCode {
     let mut envelope = Envelope::new(params);
     envelope.arm();
 
-    let regulator = PitchRegulator::new(KP_A_PER_RAD, KD_A_PER_RAD_S);
+    let regulator = PitchRegulator::new(KP_NM_PER_RAD, KD_NM_PER_RAD_S);
     let mut estimator = ComplementaryFilter::with_trust_band(ESTIMATOR_TAU_S, 0.0);
     let mut wheel_accel = WheelAccelEstimator::new(WHEEL_ACCEL_TAU_S);
 
@@ -172,9 +179,12 @@ fn main() -> ExitCode {
         let wheel_rate_rad_s = obs.erpm * RAD_S_PER_ERPM;
         let aiding = wheel_accel.update(wheel_rate_rad_s * DEFAULT_R_EFF_M, dt_s as f32);
         let attitude = estimator.update(std::slice::from_ref(&sample), aiding);
-        let proposed = regulator.update(attitude.pitch_rad, attitude.pitch_rate_rad_s, 0.0);
+        let proposed_torque_nm =
+            regulator.update(attitude.pitch_rad, attitude.pitch_rate_rad_s, 0.0);
+        // The single kt division -- the actuation boundary (issue #137).
+        let proposed_amps = proposed_torque_nm / KT_NM_PER_A;
         let (bounded_cmd, _envelope_sat) =
-            envelope.apply(Command::MotorCurrent { amps: proposed }, Faults::NONE);
+            envelope.apply(Command::MotorCurrent { amps: proposed_amps }, Faults::NONE);
 
         if let Err(e) = backend.apply(&bounded_cmd) {
             eprintln!("impulse-response-rust: apply failed: {e:?}");
--- a/crates/board-types/src/lib.rs
+++ b/crates/board-types/src/lib.rs
@@ -328,11 +328,24 @@ pub enum IoError {
 /// equal the hardware backend's, from this same struct** — otherwise a CI
 /// margin gate is not comparable to hardware, which breaks the whole "sim as a
 /// margin instrument" bet.
+///
+/// `kp`/`kd` are **N·m-denominated** (issue #137) — a property of the plant
+/// (its inertia and restoring stiffness), not of the motor. `kt_nm_per_a` is
+/// the one place a fitted motor torque constant crosses this seam: it is used
+/// only to convert a torque command to the amps the drive accepts and to turn
+/// `max_current_a` into a torque ceiling ([`Params::tau_max_nm`]), never inside
+/// the control law itself.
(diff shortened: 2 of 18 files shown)


MikePaNtZ:
Addressed the ABI version defect:

1. \`ABI_VERSION\` bumped 1 -> 2.
2. \`ObParamsV1\` -> \`ObParamsV2\` (Rust struct **and** the Python ctypes mirror in \`sim/scenarios/rust_controller.py\`) -- took the rename, since an unversioned-looking name is exactly how the next field change gets made against a struct that isn't actually stable.
3. \`tests/test_closed_loop.py:77\` now asserts \`abi_version == 2\`.

No gain change, no re-tune, float divergence untouched, as instructed. Full \`cargo test --workspace\`, \`pytest tests/\` (273 passed, 2 pre-existing xfailed), \`cargo fmt --check\`, \`cargo clippy -D warnings\`, and \`cargo run -p xtask -- gate\` all clean. Pushed to the same branch, no merge, no auto-merge.

MikePaNtZ:
Reviewed as COO. **Merging.** Verified independently rather than from the summary:

| | |
|---|---|
| `ABI_VERSION` | `1` → **`2`** |
| `test_abi_version_is_the_one_this_glue_was_written_against` | asserts **`2`** |
| `ObParamsV2` in the Rust FFI | 8 references |
| `ObParamsV2` in the Python ctypes mirror | 5 references |
| Stale `ObParamsV1` in live code | **none** — the only two remaining are the comments explaining the bump, which is exactly where that name should survive |

## You were right to take the rename, and your reasoning is the better one

I offered you the option to skip it. You didn't, and the justification you gave is sharper than mine:

> an unversioned-looking name is exactly how the next field change gets made against a struct that isn't actually stable

That is the real argument. Mine was "the name encodes the version" — a tidiness point. Yours is that leaving it would **reproduce the same defect class later**, which is the difference between cosmetics and a fix.

Also good: the doc comment now states the guard's actual **scope** — a size check proves byte count, not meaning. That distinction is the whole reason the version field has to move independently, and writing it down where the next person edits the struct is worth more than the bump itself.

## On the float divergence — accepted, and the bar was mine

Restating for the record, because it will be read later: **the "numerically identical" criterion I set was impossible to meet.** `kp_nm = fl(kp_a × kt)` in f32, then `amps = fl(torque / kt)`, cannot recover the lost bits. The divergence is *inherent to denominating the law in torque*; the only way to be bit-exact is to not make the change.

The bar you actually met — and the one I should have written — is **mathematically identical, with the only permitted difference being IEEE-754 rounding from the added conversion, shown to be orders below every pinned threshold.** Verified: worst case ~0.01% on cascade travel, 1.2e-5 m against a 0.14 m margin to its gate.

And you established it the right way — **building the old library from `origin/master` and comparing directly**, rather than trusting the docstring numbers, which you found were stale. Comparing against a real baseline instead of a recorded one is what makes the claim checkable at all.

## What this unblocks

`kt` no longer reaches the loop gain. A wrong `kt` is now a **headroom** error rather than a **gain** error — bounded and survivable, instead of dropping the board. That removes one of #132's two limbs outright, and every downstream number stops silently inheriting the 0.7 placeholder.

The remaining limb — 2.9 dB of downward gain margin against plant uncertainty in `m`, `d`, `I` — stands on its own and needs no placeholder to justify the retune. Which is the cleanest possible footing for #132.

Scope discipline held throughout: no gains touched, no retune, `plant.py`'s `kt` left alone, `analyse_delay_budget.py` deliberately untouched.

PR #146, opened 2026-08-01
ops(ci): one list of sim test binaries, shared by both jobs that need them

### LINKED ISSUE #124: sim and publish-sim-artifact duplicate every build step, and drift apart every time a binary is added — three for three
**Filed after the third instance in one day.** Patching the fourth is not the fix.

## The pattern

`sim` and `publish-sim-artifact` **both run the test suite** — the latter indirectly, because `scripts/emit_claims.py` shells out to `pytest` to build the claims manifest. So both jobs need every artifact the suite loads, and their build steps are maintained by hand in two places.

Every time a new binary is added, `sim` gets the build step and `publish-sim-artifact` does not:

| Added by | Binary | Result |
|---|---|---|
| #117 (I1b) | `plant-replay` | `publish-sim-artifact` failed |
| #120 (I1c) | `impulse-response-rust` | failed again — **on master, three runs, unnoticed** |
| #123 | (the fix for the above) | patched the symptom |

**Three for three.** The next binary makes it four.

## Why it goes unnoticed, which is the worse half

`publish-sim-artifact` is **deliberately not a required status check** — and that is correct, for a stated reason: it renders through an offscreen GL stack that must never block a merge whose physics passed.

But the consequence is that when it breaks, **merges keep flowing and nothing interrupts anyone.** It broke on #117, I merged. It broke on #120, I merged. It then failed on three consecutive pushes to master and was found only because someone happened to be watching an unrelated PR's checks.

A job that can fail indefinitely without interrupting anyone is a job whose failures are free — which is the same disease as a check that reports green while enforcing nothing, wearing the opposite costume.

## Options

1. **Share the build steps** — a composite action, or a single `make ci-artifacts` target both jobs call. One place to edit; drift becomes impossible rather than merely unlikely.
2. **Stop `publish-sim-artifact` running the suite at all.** `sim` already ran it, green, on the same commit. If the claims manifest is the only reason, emit it from the `sim` job and pass it forward as a workflow artifact. This is the option that most reduces total work.
3. **Surface the failure** — e.g. `ops/inbox.sh` already reports red PRs; extend it to report a red **non-required** job on master, so "free to fail" stops meaning "free to ignore". Complementary to 1 or 2, not a substitute.

**Recommendation: 2, with 3 alongside.** The duplication exists to serve a job that should not have needed the suite in the first place, and option 3 is what stops the *next* silently-failing job — of which there will be one — from running for days.

## Acceptance criteria

1. Adding a new binary to the workspace requires editing **one** place for CI to keep working. Demonstrate it: add a trivial binary, show both jobs still green, remove it.
2. `publish-sim-artifact` either no longer runs the test suite, or runs it through the identical shared definition `sim` uses.
3. A red non-required job on `master` becomes **visible** without someone happening to look at an unrelated PR.
4. The offscreen-GL rationale for it staying non-required is preserved. **Do not solve this by making it required** — that reintroduces exactly the failure the comment warns about.

**Owner note:** `.github/workflows/ci.yml` is Senior Controls'; `ops/inbox.sh` is the COO's. If option 3 is taken, that half is mine — say so and I will do it in the same pass.

MikePaNtZ:
Closes #124.

## The problem

`sim` and `publish-sim-artifact` **both run the pytest suite** — the latter indirectly, because `scripts/emit_claims.py` shells out to `pytest` to build the claims manifest. Both therefore need every Rust binary the suite loads, and both maintained that list **by hand, in two places**.

It drifted three times in one day, always the same direction:

| Added by | Binary | Result |
|---|---|---|
| #117 (I1b) | `plant-replay` | `publish-sim-artifact` failed |
| #120 (I1c) | `impulse-response-rust` | failed again — **on master, three consecutive pushes, unnoticed** |
| #123 | (the fix for the above) | patched the symptom |

**Patching the fourth instance is not the fix.**

It went unnoticed because `publish-sim-artifact` is deliberately **not** a required status check — correctly, since it renders through an offscreen GL stack that must never block a merge whose physics passed. But a job that can fail indefinitely without interrupting anyone is a job whose failures are free.

## The fix — option 1 of the three offered

Both jobs now call **`scripts/ci_build_sim_binaries.sh`**. One place to add a binary; drift becomes impossible rather than merely unlikely.

The script also guards the *other* way this list goes stale: **`cargo build` succeeding says nothing about the artifact landing under the name the tests look for.** A rename would still produce a green build and a confusing pytest error 40 minutes later. So it verifies each artifact exists in `target/release` and fails there, readably:

```
ci_build_sim_binaries.sh: built without error, but these artifacts are
not where the tests look for them: plant-replay
A binary was probably renamed. Update this script AND the test that
shells out to it -- they are the same list in two places by necessity.
```

**Verified by deleting a binary and confirming the guard exits 1**, not just by reading it.

`scripts/` is Senior Controls' turf and `.github/workflows/ci.yml` is too (`CODEOWNERS`), so this stays entirely in-lane — a composite action under `.github/actions/` would have been COO turf.

## Why not the other two options

- **Not option 2** (stop `publish-sim-artifact` running the suite, pass the manifest forward as a workflow artifact). It genuinely reduces the most total work, but it changes *where the claims manifest is produced and what it is derived from* — a bigger change to the artifact pipeline than this drift warrants, and `docs/design-claims-manifest.md` describes that pipeline. Worth doing separately if the duplicated *test run* (not just the build steps) becomes the cost that matters.
- **Not option 3** (surface red non-required jobs on master via `ops/inbox.sh`). Complementary rather than a substitute, and `ops/` is the COO's turf — it belongs in their lane, and is arguably the more valuable half, since it fixes "free to fail means free to ignore" in general rather than for this one job.

## Verification

- `bash -n` clean; script runs green locally and reports all three artifacts present.
- Guard failure path exercised deliberately (see above).
- `.github/workflows/ci.yml` parses, and both jobs now show a single `Build the sim test binaries` step.
- `policy_check.py` — all hard checks pass.

The real proof is this PR's own `publish-sim-artifact` run, which exercises the changed job end to end.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,25 +176,12 @@ jobs:
       - name: Cache cargo
         uses: Swatinem/rust-cache@v2
 
-      # The closed-loop gate loads the REAL control law over the C ABI, so the
-      # cdylib has to exist before pytest runs. If it is missing those tests
-      # FAIL rather than skip -- a closed-loop test that passes because the
-      # controller was absent would be worse than no test at all.
-      - name: Build the control library
-        run: cargo build --release -p control-ffi
-
-      # I1b (issue #106): tests/test_rust_python_plant_replay_equivalence.py shells out to this
-      # binary to compare the Rust-hosted plant against the Python-hosted one
-      # it runs in-process. Same "fail rather than skip" rule as above.
-      - name: Build the plant-mujoco replay binary
-        run: cargo build --release -p plant-mujoco --bin plant-replay
-
-      # I1c (issue #107, AC6): tests/test_rust_hosted_impulse_response.py
-      # shells out to this binary to compare the Rust-hosted closed loop
-      # (hal + sim-backend's real plant) against the Python-hosted one. Same
-      # "fail rather than skip" rule as above.
-      - name: Build the Rust-hosted impulse-response binary
-        run: cargo build --release -p board-app-driverless --bin impulse-response-rust
+      # Every Rust artifact the suite loads or shells out to. Shared with
+      # `publish-sim-artifact` below, which runs the same suite indirectly via
+      # `scripts/emit_claims.py` -- see the script's header for why these two
+      # lists must not be maintained separately (issue #124).
+      - name: Build the sim test binaries
+        run: ./scripts/ci_build_sim_binaries.sh
 
       # ADR-0011: tests/test_cmd_envelope_reserve.py shells out to this binary
       # for the command-envelope-reserve acceptance matrix. Same "fail rather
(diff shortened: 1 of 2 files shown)


MikePaNtZ:
Closing #148 in favor of this one — the artifact-existence guard is worth keeping and I don't have a reason to prefer the python/bash split otherwise. Clear to re-queue.

---

MikePaNtZ:
Red on `doc-drift`, and the fix is one line in a commit message — not a code change:

```
DOC-OK: <reason>
```

at the **start of a line** in any commit on this branch. Both hits are legitimate but immaterial: `docs/design-claims-manifest.md` and `docs/web-artifact-pipeline.md` declare `covers: .github/workflows/ci.yml`, and this PR changes that file. Neither doc describes the test-binary list you are consolidating.

I cannot add it for you — CI reads commit messages, so it has to be in your history, not a PR-body edit.

**Worth flagging as a pattern rather than a nuisance:** that is at least the fourth ci.yml change today dragged into doc-drift by those two manifests, and both docs are mine. A manifest that fires on every change to a shared file, and is overridden every time, is on its way to becoming decorative — the same argument I accepted for the role-log heuristic (#131) and for my own session-prep doc (#144), which I had to strip a manifest from an hour after adding it.

Recording it in my context file as an open thread: those two manifests want narrowing to the parts of `ci.yml` they actually describe, or removing. **Your PR is not the place to fix it** — add the override and land.

Issue #207, opened 2026-08-02
Pin the estimator trim and the reserve's derivation -- ADR-0011 (f1)/(f2), the controls half of clearing the hold


MikePaNtZ:
## Summary

ADR-0011's second ratification replaced criterion (f) with **freeze-and-pin**. This is that work. It is the controls half of what now clears the hold.

The board currently survives full stick because the attitude estimator reads **apparent vertical**, which is `atan(a/g)` — the exact lean the acceleration physically requires. That is the textbook balance-vehicle behaviour, arrived at by accident rather than by design. **It is not versioned, not asserted anywhere, and nothing would notice if it moved.** Until it is pinned, the launch is standing on luck with a good explanation.

## The identity, so nobody re-derives it wrongly

- ADR-0011 lean sensitivity: **5.84° per m/s²**, fixed by geometry.
- One radian per g: **57.2958 / 9.80665 = 5.8425° per m/s²**.
- Measured `est − truth` residual slope: **~5.8° per m/s²**.

Same number, same physics. Feeding the regulator MuJoCo truth does **not** de-bias it — it deletes the acceleration reference and leaves a pitch-only regulator, which is why it inverts at every stick fraction down to 0.05.

### (f1) Pin the residual
Regression-test that the `est − truth` residual slope is **≈ 1 rad/g** and its static offset sits inside a pinned band. Any estimator or tuning change that moves the trim must **break CI**, not silently re-flip the board.

`test_the_estimator_residual_is_specific_force_not_a_static_bias` in `tests/test_cmd_envelope_reserve.py` already measures the slope — it needs to become an assertion with a band, not just a characterisation.

### (f2) Pin the reserve's derivation, not just its value
`CMD_ENVELOPE_RESERVE = 0.80` was derived from a peak-demand slope of **41.97 A/unit measured at the current operating trim**. So the constant is **trim-derived, not geometry-derived** — and any doc comment implying otherwise should be corrected, because that framing is a rationalisation.

Assert the measured slope in the harness so the constant goes stale loudly rather than quietly if the trim shifts.

### Incline tolerance sweep
Measure the actual authored-terrain incline the board tolerates. A static slope is an effective static pitch disturbance, and the board inverts at ±0.5° of static pitch error and survives ±0.25°, so **the predicted incline tolerance is under 0.5°** — but predicted is not measured.

This number is a hard input to the world-authoring constraint, which cannot be written as a checkable asset rule until it exists.

## Explicitly NOT in scope

[...]

PR #214, opened 2026-08-02
DO NOT MERGE -- spike: prove the ADR-0011 (f1) trim pin actually fires (#207)

MikePaNtZ:
**Do not merge. Close once CI reports.**

ADR-0011 (f1)'s acceptance criterion for issue #207 is explicit that reading
the test is not evidence it works:

> A retune that shifts the estimator trim outside the pinned band **fails
> CI**. Demonstrate it by moving the trim on purpose and watching it go red.

This is that demonstration. On top of `feat/controls/pin-trim-and-reserve`
(which adds the pin), one constant is moved:

`ACCEL_FF_GAIN_M_S2_PER_A` **0.0584 -> 0.0650**

That is a realistic retune rather than a contrived one -- the constant is
`kt / (r_eff * total ridden mass)`, so re-measuring rider mass moves it. It
shifts the settled `est - truth` trim from **-2.501 deg to -2.690 deg**,
outside the pinned +-0.10 deg band.

Expected result: the `sim` check fails on
`test_f1_the_estimator_trim_is_pinned_so_a_retune_cannot_move_it_silently`.

### DIFF
--- a/crates/plant-mujoco/src/lib.rs
+++ b/crates/plant-mujoco/src/lib.rs
@@ -46,6 +46,8 @@ extern "C" {
     fn plant_mujoco_get_qpos(data: *mut c_void, out: *mut f64, n: c_int);
     fn plant_mujoco_get_qvel(data: *mut c_void, out: *mut f64, n: c_int);
     fn plant_mujoco_timestep(model: *mut c_void) -> f64;
+    fn plant_mujoco_get_gravity(model: *mut c_void, out: *mut f64);
+    fn plant_mujoco_set_gravity(model: *mut c_void, g: *const f64);
     fn plant_mujoco_forward(model: *mut c_void, data: *mut c_void);
     fn plant_mujoco_sensor_id(model: *mut c_void, name: *const c_char) -> c_int;
     fn plant_mujoco_sensor_adr(model: *mut c_void, sensor_id: c_int) -> c_int;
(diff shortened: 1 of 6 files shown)


MikePaNtZ:
## Result: the pin fires. Closing.

`sim` **failed**, on exactly the intended test:

```
FAILED tests/test_cmd_envelope_reserve.py::
  test_f1_the_estimator_trim_is_pinned_so_a_retune_cannot_move_it_silently

AssertionError: the estimator trim has moved to -2.6896 deg from the pinned
-2.501 deg. [...] Re-derive the reserve from a re-measured slope; do not
re-baseline this number to make the build green.
assert 0.18861941897559564 < 0.1
```

Run: https://github.com/MikePaNtZ/overboard/actions/runs/30762664302/job/91535879744

Two things worth recording beyond the pass/fail.

**The measurement is bit-identical across platforms.** The GitHub `ubuntu`
runner measured `-2.6896194189755955 deg`; the same change measured
`-2.6896 deg` locally on macOS/arm64. That was the open risk in a band as
tight as +-0.10 deg on a 2.5 deg quantity, and it is answered: the band cannot
produce a platform-dependent false positive.

**A trim move of this size is not a subtle event.** Four other assertions in
the same file went red alongside it -- the peak-demand slope fell to 29.20
A/unit against a documented 42.03, the unshaped positive control stopped
inverting the board, and the loss-of-authority warning stopped firing at all.
That is the point of (f1): the trim is load-bearing for most of what
ADR-0011 measures, and until this pin existed a change of this size could
have landed with the suite still green on the numbers that mattered.

Closing without merging. The pin itself is in #215.

Issue #222, opened 2026-08-05
Pin the bootloader EEPROM version in pins.env — it is an unpinned input to every Stage-0B latency number


MikePaNtZ:
Surfaced while writing `docs/runbook-pi-first-boot.md` (#221).

`scripts/pi/pins.env` pins the kernel on the grounds that it "is the single variable that invalidates every latency number Stage 0B produces" and that "two numbers taken across a kernel change are not comparable, and nothing in the data says so."

The same argument applies to the **bootloader EEPROM**, and it is currently unpinned.

## Why it is different from every other pin

The bootloader lives in **EEPROM on the Pi itself, not on the SD card**. So:

- Flashing a fresh, fully-pinned image does **not** reset it.
- The provenance manifest that ships beside the image cannot capture it — the manifest describes the card, and this is not on the card.
- AC-11's restore test ("flash the published artefact to a blank card and boot it") passes identically on two Pis with different bootloader versions.

It is therefore the one input to the Stage-0B go/no-go number that our entire pinning discipline structurally cannot see.

## Proposal

1. Add `BOOTLOADER_VERSION` to `pins.env` with the same "held, not tracked" comment the kernel pin carries.
2. Have the Stage-0B run record `vcgencmd bootloader_version` into the run log, alongside `git_sha` — so a number is attributable even if the pin drifts.
3. Decide whether AC-11 should assert it. Leaning yes, but it is a genuine tradeoff: asserting it means a Pi with a newer bootloader fails the restore test, which may be the wrong failure mode for a hobbyist-facing public image.

Item 3 is the part worth a considered call rather than a default. Not urgent — nothing measures latency until #182's image lands — but it should be settled before AC-5/AC-6 numbers are recorded, not after.

## Owner

Senior Controls (same lane as #182).

MikePaNtZ:
## Observed values from the CEO's Pi 5 Rev 1.1, 2026-08-05

First hardware data for this issue, captured via `docs/runbook-pi-first-boot.md` §4.

```
version    086b83e3332dfc8927c56762771d082f3077a1ae (release)
timestamp  1779807685        # firmware build: 2026-05-26T15:01:25Z
update-time 1785906962       # written to this board: 2026-08-05
capabilities 0x0000007f
BOOTLOADER: up to date  (CURRENT == LATEST after `-a`)
RELEASE: default (/usr/lib/firmware/raspberrypi/bootloader-2712/default)
```

## This changes the proposal: pin the channel, not just the hash

[...]

Issue #182, opened 2026-08-02
Stage-0B implementation: build the Pi image, wire the CAN path, and land the sim-in-the-loop self-test


MikePaNtZ:
**The design was approved and the implementation issue was never opened.** #32 closed with the design half (#51: `docs/design-pi-image-stage0b.md` + reference + verification log, 14 numeric ACs). That doc says implementation is "a separate issue, opened only after COO approval is recorded." It was not. Consequence, observable in the tree today:

- No `pi/`, no `scripts/pi/`.
- `.github/workflows/` is `ci.yml` + `policy.yml` — no image build job.
- Releases are `sim-latest` / `sim-preview*` / `sim-experiments`. No image artefact.
- The **D3 spike** — the half-day gate on the builder choice — has never been run.

What *did* ship from the workstream: `can-harness` (#52) and `scripts/stage0b_runbook.py` (#27). The vehicle they were built for does not exist. A Raspberry Pi 5 arrives **2026-08-03**.

---

## Decisions taken 2026-08-02 (CEO), before work starts

| | Decision | Effect |
|---|---|---|
| **Repo boundary** | **D1 stands: `pi/` in this repo.** A split was considered and declined | The runtime contract keeps its only end-to-end CI. A separate repo bought none of the wanted features |
| **Rebuild model** | **Split: image pinned, binary hot** | Image rebuilds only on `pi/**` or a kernel-pin change. Controller binary builds on every `master` merge to a rolling `controller-latest`. **Rejected: rebuilding the image per controller merge** — `archive.raspberrypi.com` has no snapshot service, so every rebuild drifts the kernel and package set, and no two latency numbers stay comparable |
| **Target** | Raspberry Pi 5 | No `rpi-2712-rt` flavour exists, so RT means the 4K-page v8 kernel and no `NO_HZ_FULL` (design §4). **Q12 is live** |
| **Visibility** | Public release, credential-free image | Consistent with the existing repo rule; makes I4 non-negotiable |

---

## Already landed

**#181 — `overboard-loop-profile`.** AC-5's instrument, and AC-12 coverage for it. Runs today with no Pi and no image. Its first finding sets the frame for everything below: **compute p99.9 was 0.7 µs against a 2 ms period.** The control law is ~0.03% of the budget. If the Pi cannot hold 2 ms, it will not be because of our arithmetic — this is a scheduler and transport problem end to end.

---

### I0 — the D3 spike *(gate; half a day; nothing below starts until this is green)*
[...]

PR #223, opened 2026-08-05
fix(controls): the non-RT baseline moves four variables, not two — and the EEPROM pin needs its channel (#182)

MikePaNtZ:
First-boot findings from the CEO's Pi 5 Rev 1.1, run through `docs/runbook-pi-first-boot.md` (#221) on 2026-08-05. Two corrections, both to how the Stage-0B data will be read rather than to the procedure.

## The baseline caveat undercounted its confounds

The merged doc said the baseline-to-image delta moves **two** variables. It moves **four**. Stock reported `6.18.34+rpt-rpi-2712`, `PREEMPT`, on the full desktop image:

| | Baseline (stock) | Stage-0B image |
|---|---|---|
| Preemption | `PREEMPT` | `PREEMPT_RT` |
| Flavour / page size | `rpi-2712`, 16K | `rpi-v8`, 4K |
| Kernel line | **6.18.34** | **6.12.75** |
| Userspace | full desktop | `base-slim`, headless |

The kernel-line row was the miss. Stock ships 6.18.34 — **the exact version `pins.env` predicted the floating metapackage would resolve to**, now confirmed on hardware — while the pin holds 6.12 for the `mcp251xfd` fix. That is two different kernels, not one kernel built two ways.

Why it matters: as written, the doc invited a reading the data cannot support. The corrected text states plainly that this comparison **cannot** support "PREEMPT_RT bought us X µs", so that claim cannot reach the AC-5 record on the strength of a floor measurement. Capturing the baseline is still worth doing; over-reading it is the risk.

## The EEPROM record needs the channel, not just the hash

§4 now records three fields. Observed: version `086b83e3332dfc8927c56762771d082f3077a1ae`, build timestamp `1779807685` (2026-05-26T15:01:25Z), release channel **`default`** (not `latest`). `rpi-eeprom-update` resolves "latest" differently per channel, so a hash without its channel is not reproducible. Fed back into #222.

The Pi reported `BOOTLOADER: up to date`, CURRENT == LATEST, after `-a` applied.

## Not changed

No procedure step, no ordering, no pass/fail threshold. Manifest reconciled; `policy` passes.

### DIFF
--- a/docs/runbook-pi-first-boot.md
+++ b/docs/runbook-pi-first-boot.md
@@ -5,7 +5,7 @@ covers:
   - scripts/pi/flash_pi.sh
   - scripts/pi/pins.env
   - crates/loop-profiler/src/lib.rs
-reconciled: 1618e9f
+reconciled: 725c00c
 -->
 
 **Owned by: Senior Controls.** **Executed by: the CEO**, at a desk, with a monitor and a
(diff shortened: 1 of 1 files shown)


MikePaNtZ:
Consolidated into #259, rebuilt on current master. This branch went DIRTY against the flash_pi.sh and runbook changes that landed after it.

PR #224, opened 2026-08-05
docs(controls): Q12 is closed — Pi 5 firmware honours kernel=kernel8.img, measured (#182)

MikePaNtZ:
Q12 has sat in the open-question ledger since the design landed, deferred to first boot and flagged as "the difference between a card that boots and one that does not." The CEO's Pi 5 arrived today and answered it in one reboot.

## The observation

Pi 5 Rev 1.1, stock Raspberry Pi OS, via `runbook-pi-first-boot.md` §5. Appending `kernel=kernel8.img` to `config.txt`:

```
before:  6.18.34+rpt-rpi-2712
after:   6.18.34+rpt-rpi-v8
```

It boots, the flavour changes, and `auto_initramfs=1` resolved the matching initramfs with no help. Firmware auto-selection **is** overridable by the `kernel=` directive on a Pi 5.

**This confirms the image rather than changing it.** `device/overboard-pi5/config.txt` already ships `kernel=kernel8.img` under `[pi5]`; that line now has a hardware observation behind it instead of an assumption.

## What it does not prove — stated so the ledger stays honest

Two things remain, and neither is closed by this:

1. **The pinned RT package may not install as `kernel8.img`.** This test ran the 6.18 non-RT v8 kernel. `kernel=` names a *file*, so the question that actually gates our card is whether `linux-image-6.12.75+rpt-rpi-v8-rt` produces that filename. Today it rests on a code comment citing the spike's build output. `verify_kernel.sh` asserts modules and the DTB but not the boot-image filename — adding `kernel8.img` to `KERNEL_REQUIRED_FILES` would close it in CI against the real `.deb`, no hardware needed. Suggested to the I1 session on #182 rather than done here, since that branch is live.
2. **Q11 is untouched.** Whether the packaged RT kernel carries a Pi-5-specific regression the 2712 flavour does not is a different question, still needs the image, still open.

## Scope

One ledger row. No decision, threshold or procedure changes. `policy` passes.


MikePaNtZ:
Superseded by #257, which records both halves of the Q12 answer. This PR captured only the mechanism, and its stated residual — 'does not prove the pinned RT package installs itself as kernel8.img' — subsequently fired: the package installs as kernel8_rt.img and the first real card would not have booted. Leaving this open would put a known-stale row in the ledger.

PR #225, opened 2026-08-05
docs(controls): §5's fail branch read as a test step — split it into an explicit recovery section (#182)

MikePaNtZ:
Direct feedback from the CEO mid-run, on the day the runbook was first executed: he reached §5's `**Fail (does not boot):**` paragraph **after passing** and asked for step-by-step help performing it — reading a recovery procedure as a further step of the test.

That is a doc defect, not a misread. The paragraph opened with an instruction ("put the card in the host laptop and restore…") before establishing that it only applies to a failure that, in his case, had not occurred. Compressing a whole repair procedure into one dense sentence inside the pass/fail block invited exactly that reading.

## Changes

- The fail branch now states the **observation** ("the Pi does not boot — no login prompt, no SSH") and defers the procedure to a new **§5.1**, which opens with "**On a pass, skip this section entirely.**"
- §5.1 leads with the case that actually applies most of the time: **if the Pi still boots, no card reader is needed** — three commands over a normal shell. The card-reader path is explicitly scoped to "no way in at all".
- The card path is seven numbered steps, with the two failure modes that cost real time called out: **TextEdit silently saving rich text** (use `nano`), and **pulling the card without `diskutil eject`**, which turns a config problem into a re-flash. Also notes that only `/Volumes/bootfs` appears on macOS and that this is expected, since a missing rootfs looks alarming.
- Adds the thing the original omitted entirely: **restoring after a pass is optional.** The card is scratch and the image overwrites it. The only reason to return to 2712 is the second baseline — and since v8 is the flavour the image ships, a single baseline is better taken on v8.

## Scope

Wording and structure only. No procedure step, ordering, threshold or pass/fail condition changes. Manifest reconciled; `policy` passes.

### DIFF
--- a/docs/runbook-pi-first-boot.md
+++ b/docs/runbook-pi-first-boot.md
@@ -5,7 +5,7 @@ covers:
   - scripts/pi/flash_pi.sh
   - scripts/pi/pins.env
   - crates/loop-profiler/src/lib.rs
-reconciled: 1618e9f
+reconciled: 725c00c
 -->
 
 **Owned by: Senior Controls.** **Executed by: the CEO**, at a desk, with a monitor and a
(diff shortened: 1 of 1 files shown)


MikePaNtZ:
Consolidated into #259, rebuilt on current master, with three extra recovery traps learned on the second night (BOOT vs bootfs, the cmdline.txt one-line rule, and eject discipline).

PR #231, opened 2026-08-05
feat(controls): flash_pi.sh proposes the single removable disk, so a reflash is one command (#182)

MikePaNtZ:
CEO ask, from the first-boot session: make reloading the Pi "put the SD card in, run a command, done" rather than look-up-the-device-then-run-a-command.

## What changed

`--disk` becomes optional. With no `--disk`, the script proposes the removable disk **when there is exactly one**. Zero or several is an error, never a guess.

```sh
scripts/pi/flash_pi.sh                      # was: diskutil list external, then --disk /dev/diskN
```

## What deliberately did NOT change

**Detection removes the lookup step, not a safety step.** Both guards still run against whatever lands in `$DISK`, identically whether a human or the script chose it:

- Guard 1 (removable-only, internal disks refused outright) runs on the detected disk too — and on Linux detection uses `RM=1`, the same kernel flag guard 1 checks, so they cannot disagree.
- Guard 2 (**type the device identifier back**) is unchanged and still mandatory. A target the script picked deserves the *same* scrutiny as one that was typed, not less — this is the one prompt standing between a flashed card and a flashed laptop, and convenience is not a reason to weaken it.

Refusing to choose when several removable disks are present is the load-bearing part. "Pick the first one" is exactly how a script like this breaks its promise.

## Drive-by fix, caught by making it

`usage()` hard-coded `sed -n '2,30p'` to print the header. Growing the header by nine lines silently truncated `--help` — a script whose entire job is not to surprise you, quietly lying about its own interface. Now it prints from line 2 to the first line of code, so it cannot drift again.

## Testing

- `bash -n` clean; `--help` renders the full header.
- **Zero-disk path exercised on macOS** (no card inserted): errors with the manual lookup instructions.
- Both `awk` parsers verified against real `diskutil list external physical` and `lsblk -dno NAME,RM` output shapes, including the multi-disk case.
- **Not exercised: the one-disk and multi-disk live paths** — no card was in the reader at the time. The parsing is covered above, but the end-to-end single-disk run is unverified until the next real flash. Stated rather than implied.
- Avoids `mapfile`/`readarray` throughout: macOS ships bash 3.2.

### DIFF
--- a/scripts/pi/flash_pi.sh
+++ b/scripts/pi/flash_pi.sh
@@ -1,9 +1,17 @@
 #!/usr/bin/env bash
 # Flash an Overboard image to an SD card and stage credentials onto it.
 #
-#     scripts/pi/flash_pi.sh --disk /dev/disk4                 # newest release
-#     scripts/pi/flash_pi.sh --disk /dev/disk4 --image ./x.img.xz
-#     scripts/pi/flash_pi.sh --disk /dev/disk4 --dry-run       # rehearse
+#     scripts/pi/flash_pi.sh                                   # newest release
+#     scripts/pi/flash_pi.sh --disk /dev/disk4                 # explicit target
+#     scripts/pi/flash_pi.sh --image ./x.img.xz                # local image
+#     scripts/pi/flash_pi.sh --dry-run                         # rehearse
+#
+# With no --disk, the one removable disk in the machine is proposed. Zero or
+# more than one is an error, never a guess. Detection only ever REMOVES the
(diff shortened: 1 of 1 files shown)


MikePaNtZ:
Replaced by #260. Two reasons: this branch was stale against four later flash_pi.sh changes, and its detector enumerated with 'diskutil list external physical' — which does not list a MacBook's built-in SDXC reader, so it would silently have failed to find the CEO's only card. #260 filters on Removable Media, the same field the safety guard checks.

PR #246, opened 2026-08-07
fix(controls): the removable-media guard refused the CEO's actual card reader (#182)

MikePaNtZ:
**Blocking the first AC-11 restore test right now.** Found live, mid-flash.

## The bug

Guard 1 refused any disk reporting `Device Location: Internal`. A MacBook's **built-in SDXC slot reports `Internal`** — it hangs off an internal bus — while the card in it is `Removable Media: Removable`. The CEO's card:

```
Device Node:      /dev/disk7
Disk Size:        128.0 GB
Device Location:  Internal      <-- guard refused on this
Removable Media:  Removable     <-- what actually matters
Virtual:          No
```

So the script refused the only card reader he has. **That is the worst failure mode a safety guard can have:** not a missed catch, but a false refusal on the correct target, which sends the operator looking for a way around the guard. A guard that cries wolf gets disabled, and then it is not there on the day it matters.

"Which bus is the reader on" was never the question. "Is this the system disk, and is the media removable" is.

## The fix — same strictness, correct predicate

1. **Refuse the system disk by identity**, comparing `Part of Whole` against `diskutil info /`. This is the check that actually expresses the intent the old one was groping at.
2. **Removability becomes load-bearing** and must be exact (`Removable`/`Yes`). Internal NVMe reports `Fixed` and is refused here.
3. **Refuse `Virtual: Yes`** — synthesized APFS containers are not raw targets.

Still no `--force`. Guard 2 (type the identifier back) is untouched.

## Verified against the CEO's real disks

| Disk | Reports | Result |
|---|---|---|
| `/dev/disk7` (his SD card) | Removable, Internal | **ACCEPTED** — the fix |
| `/dev/disk0` (internal NVMe) | Fixed, Internal | REFUSED — *"does not report as removable media (got: 'Fixed')"* |
| `/dev/disk3` (root APFS container) | Fixed, Internal | REFUSED — *"holds the volume this Mac is booted from"* (the new check firing) |
| `/dev/disk4` (240 MB USB stick) | Removable, External | ACCEPTED |

That last row is correct and worth stating plainly: **the guards do not protect against picking the wrong removable device.** A USB stick is as valid a target as an SD card. Only size-matching distinguishes them, and that is the operator's job — now said out loud in the runbook rather than left implied.

## Knock-on: PR #231's auto-detect has the same bug

`removable_disks()` there enumerates via `diskutil list external physical`, which **will not list this card either**. #231 is still open; its detector needs the same correction before it merges, or the convenience feature will silently fail to find the CEO's only reader. Flagged, not fixed here — this PR is the one standing between him and tonight's measurement.

## Doc

The runbook told the reader to find the disk with `diskutil list external`, which cannot work on this hardware. Now says use `diskutil list`, explains why, and states the size-matching responsibility.

### DIFF
--- a/docs/runbook-pi-first-boot.md
+++ b/docs/runbook-pi-first-boot.md
@@ -5,7 +5,7 @@ covers:
   - scripts/pi/flash_pi.sh
   - scripts/pi/pins.env
   - crates/loop-profiler/src/lib.rs
-reconciled: 3251993
+reconciled: 05f5c6b
 -->
 
 **Owned by: Senior Controls.** **Executed by: the CEO**, at a desk, with a monitor and a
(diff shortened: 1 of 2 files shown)


MikePaNtZ:
Replaced by #258, rebuilt cleanly on current master. This branch went DIRTY against #236, #247 and #256, all of which touched flash_pi.sh after it was opened. Same fix, same verification, no stale conflict.

Issue #249, opened 2026-08-07
The Stage-0B image hangs on first boot, cannot retry, and ships with no route in — four defects from the first real flash


MikePaNtZ:
Four defects, all found on the first real boot of the Stage-0B image (Pi 5 Rev 1.1, 2026-08-06). **Every one was invisible without hardware** — CI was green throughout.

The headline is good and should not be lost among them: the image **boots on the pinned RT kernel**.

    Linux overboard 6.12.75+rpt-rpi-v8-rt #1 SMP PREEMPT_RT aarch64

Hostname set, user created, SSH key installed, `vcan0` up by itself. Q11 answered: no Pi-5-specific RT regression prevents boot.

---

## D1 — `overboard-firstboot` deadlocks, permanently, on every first boot

**Severity: the image is unusable as shipped.** Boot stops at `Job overboard-firstboot.service/start running` and never proceeds. `Type=oneshot` disables `TimeoutStartSec` by default, so it hangs *forever* rather than failing.

Journal shows exactly where — the log stops after the last line before it:

    overboard-firstboot: SSH password authentication disabled (key-only)

The next statement is line 97:

    systemctl enable --now ssh

`--now` starts a unit from inside the boot transaction that is currently blocked on this very unit. Circular wait.

Line 113 has the same bug in a second flavour, not reached because 97 hangs first:

    nmcli connection reload

The unit declares `Before=NetworkManager.service`, so `nmcli` tries to D-Bus-activate a service ordered *after* it.

**Fix:** `systemctl enable ssh` (no `--now`) — the symlink is enough, `multi-user.target` starts it. Replace the `nmcli` call with a no-op, or drop `Before=NetworkManager.service` and reload properly. Verified live: with both patched, the board boots to a login prompt.

**Also add `TimeoutStartSec=` to the unit.** A first-boot credential installer must fail the boot loudly, never wedge it silently.

## D2 — first boot cannot be retried, so a failure is terminal

`ConditionPathExists=/boot/firmware/overboard-userconf` makes the unit inert once that file is gone. After the deadlocked boots, the third boot logged:

    skipped, unmet condition check ConditionPathExists

So the board came up **without WiFi profiles ever installed** and no way to re-run short of re-flashing. A first-boot installer needs a success marker written *on completion*, not a trigger file that can disappear mid-failure.

## D3 — the WiFi path depends on tools the image does not contain

    WARNING: could not set regulatory domain; 5 GHz may be unavailable

[...]

PR #251, opened 2026-08-07
fix(controls): first boot no longer deadlocks, can retry, and ships a route in (#249)

MikePaNtZ:
## What

Fixes the four defects filed in #249, all found on the first real hardware boot (Pi 5 Rev 1.1, 2026-08-06) and invisible to CI because every one of them only shows up on real hardware.

## D1 — deadlock, fixed

`overboard-firstboot.service` hung forever on `systemctl enable --now ssh`: the unit runs inside the same boot transaction that `--now` tries to start ssh synchronously in — a circular wait, and `Type=oneshot` disables `TimeoutStartSec` by default so it never times out. Dropped `--now`; `enable` alone is enough, `multi-user.target` starts the unit once the transaction clears.

`nmcli connection reload` a few lines later has the identical bug in a second flavour (the unit is `Before=NetworkManager.service`, so `nmcli` tries to D-Bus-activate a service ordered to start after it) — not reached before, because D1 hung first. Removed the reload call entirely: since the unit is ordered before NetworkManager starts, NM reads the freshly-dropped connection profiles itself on its own first start. No reload was ever needed.

Also added `TimeoutStartSec=120` so a *future* hang here fails the boot loudly instead of wedging it silently, per the issue's explicit ask.

## D2 — retry, fixed

The unit gated on `ConditionPathExists=/boot/firmware/overboard-userconf`, but the script deletes that same file as its last step — so a script that died partway (or was interrupted mid-hang) looked, to the next boot, identical to one that had finished cleanly. Replaced it with a completion marker (`overboard-firstboot.done`) the script writes only once it reaches the end without error; the unit now gates on that marker's *absence*. A partial run leaves no marker and retries on the next boot; a clean run (real work or a true no-op) writes it once and stops re-running.

I did not try to re-derive exactly how the userconf marker was already gone by the third boot in the CEO's report — the fix makes that mechanism irrelevant rather than depending on identifying it.

## D3 / D4 — missing packages, fixed

`overboard-firstboot` already calls `iw reg set` as its regulatory-domain fallback and the runbook/`flash_pi.sh` promise `overboard.local`, but the image shipped none of `iw`, `rfkill`, or `avahi-daemon`. Added all three to `overboard-base.yaml`'s package list — the "ship it" option the issue itself called the better answer for a headless board with no console.

## Deliberately left out

- **The runbook's "console password is optional" line** — `docs/` is COO turf, not mine, and the issue itself says this needs correcting. Flagging it here rather than editing it.
- **Root-causing exactly why the userconf trigger file was already absent on the third boot** — the completion-marker redesign makes the retry story correct regardless of that mechanism, so it wasn't necessary to chase down, and this repo's convention is not to chase what isn't blocking.

## Verification

Hardware-in-the-loop is the only real verification for a first-boot script, and this session has none. What I could check from here:
- `bash -n` on the modified script — clean.
- `python3 scripts/pi/check_layer_headers.py` — all layer files still parse as valid DEB822.
- `python3 .github/policy_check.py` — all hard checks pass; no new advisories from this change.
- Traced the systemd unit ordering (`Before=NetworkManager.service`, transitively before `ssh.service` via `multi-user.target`) by hand against both bugs to confirm the circular-wait mechanism the issue describes, and that removing `--now`/the reload call doesn't lose any functionality NM/systemd wasn't already going to provide on its own.

Real confirmation needs another flash + boot cycle on hardware, same as the original report.

---
_Generated by [Claude Code](https://claude.ai/code/session_01L59K2L1AWibRiFWqwU3cHb)_

### DIFF
--- a/scripts/pi/image/layer/overboard-base.rootfs-overlay/etc/systemd/system/overboard-firstboot.service
+++ b/scripts/pi/image/layer/overboard-base.rootfs-overlay/etc/systemd/system/overboard-firstboot.service
@@ -18,14 +18,23 @@ RequiresMountsFor=/boot/firmware
 # but must not be RUNNING first -- the installer reloads it afterwards.
 Before=NetworkManager.service
 Wants=NetworkManager.service
-# Idempotent, but the unit still guards on the marker so a normal boot does
-# not pay for a script that will immediately no-op.
-ConditionPathExists=/boot/firmware/overboard-userconf
+# Idempotent, but the unit still guards on a marker so a normal boot does not
+# pay for a script that will immediately no-op. Gated on the ABSENCE of a
+# completion marker the script writes on success -- not on the presence of
+# the staged credential files, which the script itself deletes as its last
+# step. Gating on those made a script that died partway indistinguishable
+# from one that finished: the next boot saw the trigger file already gone
+# and never retried.
+ConditionPathExists=!/boot/firmware/overboard-firstboot.done
 
 [Service]
 Type=oneshot
 RemainAfterExit=yes
 ExecStart=/usr/local/sbin/overboard-firstboot
+# A first-boot credential installer must fail the boot loudly, never wedge
+# it silently. Type=oneshot disables TimeoutStartSec by default, so without
+# this a hang here blocks the boot transaction forever.
+TimeoutStartSec=120
 # Deliberately NOT `StandardOutput=null`: on a headless board with no user yet
 # configured, the journal is the only record of why the network did not come
 # up, and that is the failure this unit exists around.
--- a/scripts/pi/image/layer/overboard-base.rootfs-overlay/usr/local/sbin/overboard-firstboot
+++ b/scripts/pi/image/layer/overboard-base.rootfs-overlay/usr/local/sbin/overboard-firstboot
@@ -24,12 +24,20 @@ NET_SRC="$BOOT/overboard-net.d"
 NM_DIR="/etc/NetworkManager/system-connections"
 USERCONF="$BOOT/overboard-userconf"
 AUTHKEYS="$BOOT/overboard-authorized_keys"
+# Written only once this script reaches its end without error. The unit's
+# ConditionPathExists gates on this file's ABSENCE, not on USERCONF/AUTHKEYS
+# still being there -- those get shredded as the last step below, so gating
+# on them made a script that died partway (or was interrupted by a hung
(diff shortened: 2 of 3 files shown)


MikePaNtZ:
Heads-up: **#256 merged ~an hour ago and covers the same six defects.** Two sessions solved #249 in parallel — my fault for not checking for an existing branch before starting. This PR is now DIRTY against it.

Comparing the two, they agree on almost everything:

| | #251 (yours) | #256 (merged) |
|---|---|---|
| `enable --now ssh` deadlock | ✅ | ✅ |
| `nmcli` D-Bus deadlock | ✅ | ✅ |
| `TimeoutStartSec=120` | ✅ | ✅ |
| Done-marker instead of trigger-file condition | ✅ | ✅ |
| `iw`, `rfkill`, `avahi-daemon` | ✅ | ✅ |
| `video` group for `vcgencmd` | ❌ | ✅ |
| Write cfg80211 regdom directly | ❌ (ships `iw` so the existing fallback works) | ✅ (both) |

**Your marker placement is better than mine and I am porting it.** You put the done-marker on the FAT boot partition (`/boot/firmware/overboard-firstboot.done`); I put it on the rootfs (`/var/lib/overboard/firstboot-done`). Yours can be deleted **from a Mac with the card in a reader**, so first boot can be forced to re-run without booting the board at all. Given that the failure this marker exists to survive is *"the board will not boot"*, a recovery lever reachable only by booting the board is the wrong side of the partition. That is a better call than mine and it took a second session to see it.

Everything else in #256 is equivalent or a superset, so I would suggest closing this rather than rebasing it — but it is your branch and your call.

Two things worth knowing that came out of the hardware session and are now in master: the RT kernel installs as `kernel8_rt.img`, not `kernel8.img` (#247, would not have booted at all), and **AC-5 passes** — p99.9 72 µs, max 113 µs (#254).

MikePaNtZ:
Confirmed independently before closing — diffed #256 against this branch. It's a strict superset: same five fixes here (both deadlocks, `TimeoutStartSec=120`, the done-marker redesign, `iw`/`rfkill`/`avahi-daemon`), plus the `video` group for `vcgencmd` and the direct cfg80211 regdom write that this PR didn't cover, plus the doc/`flash_pi.sh` wording fixes that were out of scope here (COO turf).

Glad the boot-partition marker placement was worth porting — closing this in favor of #256 rather than rebasing, since there's nothing left here it doesn't already have.

---

Issue #261, opened 2026-08-08
CI can prove the image is well-formed but not that it boots — add a QEMU first-boot gate


MikePaNtZ:
Every one of the six defects in #249 was invisible to CI. `pi-image` was green on the exact commit that produced a card which **hung forever on first boot**.

That is the gap worth closing. We can currently verify a great deal about the image and nothing at all about whether it works.

## What CI checks today

| Check | Catches |
|---|---|
| `verify-pins` | every package pin resolves |
| `verify-kernel` | required modules, DTB and config flags are in the .deb |
| `check-pin` | the kernel is held, not floating |
| `build-image` | the image assembles |
| **#247's new assertion** | `config.txt` names a kernel file that exists |

All useful. **None of them boots anything.**

## What only a boot catches

Confirmed on hardware 2026-08-06, every one of these with a green build:

- `overboard-firstboot` deadlocking forever on `systemctl enable --now ssh`
- `Type=oneshot` disabling the start timeout, so it wedges rather than fails
- `ConditionPathExists` on the trigger file making a failed first boot terminal
- `raspi-config` / `iw` / `rfkill` being called but not installed
- no `avahi-daemon`, so `overboard.local` can never resolve
- the created user missing the `video` group, so `vcgencmd` fails

Six defects, one hardware session, zero CI signal.

## Proposal — boot the image in QEMU and assert it reaches a login prompt

`qemu-system-aarch64` can boot the built artefact with the kernel and DTB extracted from the boot partition. It does not need a Pi, and it runs on the same `ubuntu-24.04-arm` runner the image is built on.

Minimum viable check, which would have caught **all six**:

1. Boot the image with a synthetic `overboard-userconf` and `overboard-net.d` staged on the boot partition, exactly as `flash_pi.sh` does.
2. Assert boot reaches `multi-user.target` **within a timeout**. This alone catches the deadlock — the single highest-value assertion available.
3. Assert `systemctl is-active ssh` and `systemctl is-system-running` is not `degraded`.
4. Assert the created user exists and is in `video`.
5. Assert `getent hosts` finds the avahi stack, and that `iw`/`rfkill` are present.
6. Assert `/var/lib/overboard/firstboot-done` exists — first boot actually completed.
7. Reboot once and assert first-boot is skipped and the system still reaches `multi-user.target`.

## Known limits, stated up front

[...]

PR #263, opened 2026-08-08
fix(controls): QEMU boot-test harness for the Pi image -- root-device recognition still open (#261)

MikePaNtZ:
## Status (updated after CI iteration — see PR comment for the full account)

**The infrastructure works. The boot itself does not yet reach userspace.** Three independent, evidence-backed attempts at presenting a root disk to this kernel's initramfs (`virtio-blk` MMIO, `virtio-blk` PCI, `ahci`+`ide-hd` PCI) all failed identically — the initramfs never finds its root device and drops to its own recovery shell. A diagnostic step (`lsinitramfs` module inventory, still in this PR) proved the virtio hypothesis wrong on real data rather than by further guessing; the AHCI attempt is unexplained by anything short of interactive access into that shell, which this session does not have. Full account in the PR comment thread.

**Not closing #261 with this PR** — its actual ask (a working boot gate) is not yet achieved. Leaving it open for whoever picks up the root-device question next, who inherits three ruled-out approaches and a working extraction/staging harness rather than a blank page.

The `boot-test` job is not a required check (same as every other job in `pi-image.yml`) and reports red honestly, matching this file's own `spike` job convention: red is a finding here, not a hidden failure and not something to force green with another unverified guess.

## What changed

Adds a `boot-test` job to `.github/workflows/pi-image.yml` and a new script, `scripts/pi/qemu_boot_test.sh`, that:
1. Downloads the `build-image` artefact, decompresses it, loop-mounts it.
2. Extracts the kernel (`kernel8_rt.img`) and its paired initramfs (`initramfs8_rt`, derived from `config.txt`'s `kernel=` line the same way the real firmware's `auto_initramfs=1` does) — confirmed working, every run.
3. Lists the initramfs's module inventory for virtio/common root-fs drivers — a genuine, reusable diagnostic, not just a debugging aid for this PR.
4. Stages **synthetic, hand-written** boot-partition credentials (placeholder PSK, a fake-but-plausible SSH key). Never calls `mk_boot_config.py` — that script's own docstring forbids running it in CI or from an agent.
5. Injects a CI-only systemd unit onto a throwaway loop-mounted rootfs copy that would print `QEMU-CI-ASSERT: <name>=PASS|FAIL` lines and drive a two-phase reboot check — written and believed correct, but **never yet exercised**, because boot doesn't get far enough to reach it.

None of the injected/staged content is baked into the published image — every file lives only in the ephemeral loop-mounted copy the job downloads and discards.

## Acceptance criteria (from #261's proposal) — honest state, not aspirational

1. ✅ Synthetic boot-partition config staged in the right shape.
2–7. ❌ **Not yet achieved.** All of these depend on reaching `multi-user.target`, which the boot does not currently do. Every named check in the CI-only assertion script is real code that would answer these questions correctly if the machine ever got that far — untested in anger.

## Known limits, stated up front

QEMU's `virt` machine is not a Pi 5 — no QEMU machine models the Pi 5/BCM2712 at all. RP1, real SPI, the CAN HAT, thermals and every latency number stay hardware-only (AC-5, AC-6) regardless of how the root-device question resolves.

## Deliberately left out

- **Real mDNS resolution** (issue ask 5's `getent hosts` check) — no virtual network device is attached; substituted with `systemctl is-active avahi-daemon` in the (currently unreached) assertion script.
- **A Pi-specific QEMU machine model** (`raspi4b`/BCM2711) as the next thing to try for the root-device question — flagged in the PR comment, not attempted here; it reopens the hardware-fidelity trade-off this PR originally argued against `virt`, and deserves its own deliberate pass rather than a fourth guess appended to this one.
- **Interactive serial access** to actually inspect `/proc/modules`/`dmesg` from inside the initramfs's recovery shell — the decisive way to finish this, not available to a `-serial file:` sink.

## Verification

- `bash -n scripts/pi/qemu_boot_test.sh` and the embedded assertion script — clean.
- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/pi-image.yml'))"` — clean.
- `python3 scripts/pi/check_no_secrets.py` — clean.
- `python3 .github/policy_check.py` — all hard checks pass; turf override on `.github/workflows/pi-image.yml` recognized correctly across every commit on this branch.
- **Six real CI runs against real `ubuntu-24.04-arm` hardware**, each one either fixing a concrete, evidenced defect (default-NIC romfile, missing initramfs, wrong transport family) or answering a diagnostic question. The remaining failure is the AHCI root-device gap described above.

Related: #261 (not closed by this PR).

### DIFF
--- a/.github/workflows/pi-image.yml
+++ b/.github/workflows/pi-image.yml
@@ -185,7 +185,7 @@ jobs:
       - name: Install QEMU
         run: |
           sudo apt-get update -qq
-          sudo apt-get install -y --no-install-recommends qemu-system-arm zstd xz-utils
+          sudo apt-get install -y --no-install-recommends qemu-system-arm zstd xz-utils initramfs-tools-core
 
       - name: Download the built image
         uses: actions/download-artifact@v4
--- a/.github/workflows/pi-image.yml
+++ b/.github/workflows/pi-image.yml
@@ -162,6 +162,44 @@ jobs:
           if-no-files-found: error
 
   # -------------------------------------------------------------------------
+  # Issue #261. Every one of the six defects #249 found on real hardware was
+  # invisible to every job above: the build was green on the exact commit
+  # that produced a card that hung forever on first boot. This boots the
+  # built image and asserts it actually reaches multi-user.target with
+  # userspace configured the way overboard-firstboot promises to.
+  #
+  # NOT a required check, same reasoning as the rest of this file: it needs
+  # `build-image` to have produced an artefact first, and its own boot
+  # timeout makes it a further tens of minutes on top of that. NOT a
+  # hardware-fidelity test either -- see scripts/pi/qemu_boot_test.sh's own
+  # header for exactly what it does and does not prove. AC-5/AC-6 stay
+  # hardware-only.
+  # -------------------------------------------------------------------------
+  boot-test:
+    runs-on: ubuntu-24.04-arm
+    needs: [build-image]
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install QEMU
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y --no-install-recommends qemu-system-arm zstd xz-utils
+
+      - name: Download the built image
+        uses: actions/download-artifact@v4
+        with:
+          name: overboard-image-${{ github.sha }}
+          path: /tmp/overboard-image
+
+      - name: Boot it and assert first-boot behaves
+        run: |
+          img="$(ls -1 /tmp/overboard-image/*.img.xz /tmp/overboard-image/*.img.zst /tmp/overboard-image/*.img 2>/dev/null | head -1)"
+          [ -n "$img" ] || { echo "FATAL: no image found in the downloaded artefact"; exit 1; }
+          bash scripts/pi/qemu_boot_test.sh --image "$img"
+
+  # -------------------------------------------------------------------------
   # I0 -- the D3 spike. Gates the builder choice for everything after it.
   #
   # Native arm64, no QEMU: `ubuntu-24.04-arm` hosted runners are GA and free
--- a/scripts/pi/qemu_boot_test.sh
+++ b/scripts/pi/qemu_boot_test.sh
@@ -1,0 +1,329 @@
+#!/usr/bin/env bash
+# Issue #261 -- CI can prove the built image is well-formed but never boots
+# it. Every one of the six defects #249 found on real hardware (2026-08-06)
+# was invisible to `verify-pins`/`verify-kernel`/`check-pin`/`build-image`:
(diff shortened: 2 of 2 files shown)


MikePaNtZ:
## Status: infrastructure works, the boot itself doesn't get to userspace yet — stopping here rather than continuing to guess

The pipeline through kernel/initramfs extraction and staging is solid and confirmed working on real CI (loop-mount, kernel+initramfs pairing, credential staging, unit injection all execute cleanly every run). The unresolved part is narrower and more specific than the PR description currently implies: **no generically-presented QEMU block device gets recognized as the root filesystem by this kernel's initramfs.**

Three independent, evidence-backed attempts, each ruled out by an actual CI run rather than superseded by guesswork:

1. **`virtio-blk-device` (MMIO transport)** — failed, "/dev/vda2 does not exist."
2. **`virtio-blk-pci`** — failed identically. Traced to the real cause via a diagnostic step (`lsinitramfs` module inventory): this kernel's initramfs ships **no virtio driver at all**, under any name — unsurprising, real Pi hardware has none.
3. **`ahci` + `ide-hd` (PCI)** — `ahci.ko` is confirmed present in the initramfs inventory, PCI is confirmed live on the `virt` machine (that's what the default-NIC romfile finding proved). Still failed identically: "/dev/sda2 does not exist."

All three fail at the exact same point (`local-block`'s udev-triggered device wait gives up, drops to the initramfs shell) regardless of transport, which is the kind of pattern that stops being informative to keep permuting blindly. I don't have interactive access into that shell (`-serial file:` is output-only) to go further from here, and this project's own dead-end log is explicit about the cost of guessing infra behavior round after round instead of stopping to get real data — I did that once already (the `lsinitramfs` step), and it answered the virtio question decisively but not the AHCI one.

**What's likely worth trying next, not attempted here:** a Pi-specific QEMU machine model (`raspi4b`/BCM2711, despite the SoC mismatch caveat already in this PR — the kernel may simply want Pi-native platform devices, not any generically-attached PCI/MMIO block device) or genuinely interactive serial access to inspect `/proc/modules` and `dmesg` from inside the initramfs shell. Neither is a small follow-on to this PR; both need their own dedicated pass.

**What I'm doing instead of a fourth guess:** leaving the boot-test job as-is (still not a required check, per this file's own existing convention — nothing in `pi-image.yml` blocks a merge). It will keep reporting red honestly, the same way this file's own `spike` job is designed to report red as *a finding, not a broken build*. I'm not forcing a false green, and I'm not spending more CI rounds on unverified guesses.

The value already delivered and unaffected by this: the loop-mount/extraction/staging pipeline is real, tested infrastructure, and the `lsinitramfs` diagnostic step is itself a genuine, reusable finding about this image's driver set. Whoever picks up the boot-to-multiuser piece next has three ruled-out approaches and a working harness to iterate against, instead of starting from zero.

---

Issue #228, opened 2026-08-05
test_terrain.py attributes its steady/rolling inversion to a mechanism that scenario does not use


MikePaNtZ:
## What

`tests/test_terrain.py` explains its steady-vs-rolling result by pointing at a current-proportional bias in `CommandFeedforward`'s estimator.

`sim/scenarios/terrain.py` and `sim/scenarios/hill.py` default to `control-ffi`'s `estimator_accel_aiding=1` — the `WheelAccelEstimator` wheel-odometry path. `CommandFeedforward` is not the active estimator in the scenario the docstring is describing.

So the attributed mechanism cannot be the operative one, or the docstring is describing a configuration the test does not run.

## Why it is worth fixing rather than ignoring

An explanation written next to a measurement gets read as part of the measurement. This one is currently the standing hypothesis for why the "rolling terrain is harder than steady grade" headline **inverts** on the 60 A envelope at 10% grade (at 12–14%, steady fails while rolling does not). Anyone picking that thread up starts by investigating the wrong component.

This repo has already paid for this once: ADR-0011's second ratification exists because an interpretation was carried forward while the measurement under it was sound but misattributed — *"the measurement held, the interpretation did not."*

## Asks

1. Correct the attribution, or state which configuration the claim applies to.
2. Confirm whether the steady/rolling inversion reproduces under `CommandFeedforward` as well, since that determines whether it is an estimator-path artefact or a plant-level result.

Surfaced during 60 A envelope measurement. Deliberately not root-caused there — chasing it was out of scope for a go/no-go check, and it wants its own pass.

Issue #227, opened 2026-08-05
ADR-0011's f1/f2 pin has no coverage of the wheel-odometry estimator path the scenarios actually use


MikePaNtZ:
## What

ADR-0011's freeze-and-pin criteria (f1)/(f2) — the argument the whole launch-hold exit rests on — regression-test **only** the `CommandFeedforward` estimator path. `crates/sim-host/src/host.rs` hardcodes `CommandFeedforward` and offers no wheel-odometry option.

But `sim/scenarios/hill.py` and `sim/scenarios/terrain.py` — and therefore `tests/test_terrain.py`, the suite that produced the anomaly under investigation — default to `control-ffi`'s `estimator_accel_aiding=1`, the `WheelAccelEstimator` wheel-odometry path. That path is pinned by nothing.

## Why it matters

"The estimator trim is frozen and pinned" is doing load-bearing work in ADR-0011's second ratification. As written it is a claim about one estimator path, while the scenarios that exercise the board over terrain use a different one. `control-core`'s own documentation concedes the gap: *"outdoors on a hill [CommandFeedforward] would not [win]... blending the two is the obvious next step"* — which implies the wheel-odometry path is the one closer to hardware, and it is the unpinned one.

## Also: the pin was only ever characterised on flat ground

`PINNED_TRIM_DEG = -2.501`, `PINNED_TRIM_BAND_DEG = 0.10` (`tests/test_cmd_envelope_reserve.py`) were measured on flat ground. Measured on grade, the residual runs **5×–38× outside** that band at 10–14% (both estimator paths):

| grade | incline | direction | residual (deg) | vs pinned −2.501±0.10 |
|---|---|---|---|---|
| 10% | 5.71° | downhill-fwd | −5.15 | 26× band |
| 12% | 6.84° | downhill-fwd | −5.71 | 32× band |
| 14% | 7.97° | downhill-fwd | −6.28 | 38× band |
| 10–14% | | uphill-fwd | −2.04 … −0.66 | 4.6×–18× band |

Neither the ADR nor the test states that the pin's scope is flat ground. A reader takes "frozen and pinned" as unconditional. It is not.

## Not currently a shipped-world problem

[...]

PR #264, opened 2026-08-08
fix(controls): a near-zero accelerometer reading no longer resolves to a confident 180° (#250)

### LINKED ISSUE #250: A near-zero accelerometer reading makes the estimator report 180° — atan2(0, -0), and the controller believes it
## What

The complementary-filter estimator resolves attitude from the accelerometer. When true specific force is **near zero**, that computation degenerates:

```
atan2(0, -0) = 180°
```

The estimator reports **the board is upside-down**. The controller trusts it and saturates corrective current at the full **−60 A** rail. Because the estimator's `tau = 2 s` decays slowly relative to the transient, it holds that command long enough to drive a *real* physical divergence.

Found in sim: the terrain scenario spawns out of ground contact (`data.ncon == 0`), so the accelerometer correctly reads `[0, 0, 0]` — free-fall — and the board crashes within ~0.25 s from a standing start, before any command is issued. Confirmed estimator-specific: the same plant with `use_estimator=False` (truth pitch) **survives cleanly**.

## Why this is not just a sim-spawn bug

**A real board can read near-zero specific force.** Any brief unloading does it — cresting a rise fast enough to go light, a drop off a kerb, landing, a jump. Those are ordinary riding events on the terrain this board is meant for, not exotic edge cases.

In those moments the current code will resolve an attitude from a vanishing vector, and the answer is arbitrary — dominated by whatever noise is present. In sim, ~6% of random seeds happen to produce a survivable angle and the rest produce a fatal one. That is a lottery, not an estimator.

The consequence on hardware is a full-rail current command based on a garbage attitude, at exactly the moment the wheel has little or no traction to absorb it.

## The class of bug this repo already names

`test_the_shuttle_hands_the_controller_a_live_imu`'s docstring already describes this failure class. The mechanism is known; the guard is missing.

## Asks

1. **Reject or hold the accelerometer update when `|a|` falls below a threshold** — a vanishing vector carries no direction information, so the correct response is to coast on the gyro rather than trust an arbitrary angle. Derive the threshold; do not pick it.
2. Decide the behaviour on sustained low-g (a long drop): gyro-only integration drifts, so there must be a defined recovery and probably a fault signal.
3. Consider whether this deserves a `Faults` bit — the rider and the log should both know attitude was unreliable, rather than discovering it from the outcome.
4. Add a test that drives specific force to zero and asserts the estimator does **not** invert.

## Context

Surfaced while root-causing why terrain scenarios appeared seed-fragile. The spawn defect that exposed it is filed separately; this issue is the estimator's response to a degenerate input, which is the part that would follow us onto real hardware. Related: ADR-0011's third ratification, and the freeze-and-pin construction in criteria (f1)/(f2) which assumes the estimator's residual is well-behaved.

MikePaNtZ:
## What changed

`ComplementaryFilter::accel_pitch` computes `atan2(f_x − a, −f_z)`. Fed a (near) zero specific-force vector — any brief unloading: cresting a rise, a kerb, a drop, a jump — this still returns a confident angle, most often 180° exactly (`atan2(0, -0)`), and the old estimator trusted it immediately. The controller then saturates corrective current at the full rail against an attitude that was never measured.

`crates/control-core/src/lib.rs`: added `MIN_TRUSTED_ACCEL_MAG_M_S2` (half of g — derived, not picked; see the doc comment for the full reasoning) and made `accel_trusted` gate on it **unconditionally**, independent of the existing opt-in `accel_trust_band_m_s2` disturbance-rejection knob (unchanged — still off by default, still the caller's choice). Below the floor the estimator coasts on the gyro instead of fusing a directionless vector, both in steady-state fusion and, separately, on the very first sample (the `!initialised` snap-to-accelerometer path) — so a board that spawns or resumes out of ground contact no longer initialises on the degenerate angle either.

Two new `control-core` unit tests pin the fix directly: a first-sample free-fall reading does not snap to a degenerate angle and initialises cleanly on the next trusted sample; sustained free-fall for 5 s (several τ) never converges toward 180°.

Because the floor is unconditional, it applies to every existing caller with no other code changes needed — `sim-host`'s two `ComplementaryFilter::with_trust_band(ESTIMATOR_TAU_S, 0.0)` call sites are unaffected by the literal `0.0` (that argument is the *other*, still-optional gate), and every `control-ffi` consumer (including the Python scenario harness via `rust_controller.py`) is protected without any FFI default needing to change.

## The consequence found by actually running the full suite

Re-running the whole test suite (not just `control-core`) found `tests/test_terrain.py` red: 4 of its GATE assertions — including the file's own stated headline, "a rolling profile is harder than a steady grade of the same steepness" — no longer held. Root-caused rather than reverted: the terrain scenario's dip and transitions produce genuine near-unloading moments that the *old*, buggy estimator was mishandling via this exact defect, which was (at least partly) what those assertions were actually measuring instead of what they claimed to measure.

Re-swept grade rather than guessing a replacement number (full detail, including confirmation the steady-grade ceiling is unaffected by this fix — checked against unfixed code directly via `git stash` — in the module docstring and the role log entry):

- **Headline (steady vs rolling).** No grade exists anymore where steady survives and rolling doesn't — steady's own ceiling (10.0%, unrelated to this fix) is now *below* rolling's (10.5%). Re-pinned to the real, repeatable 10.1–10.5% window where the relationship is the exact opposite of the original headline.
- **Estimator costs the envelope** (truth vs estimate). Same shape of comparison, re-derived from 10.0% to 11.0% peak grade, chosen from the middle of a measured 10.6–11.4% window (11.6% already flips back — the same non-monotonic-near-a-boundary signature issue #24 AC2's disturbance sweep already documented elsewhere in this codebase).
- **Dip vs descent estimator RMS.** Did not just shift — inverted categorically across the whole 2–9% grade sweep. Re-pinned to the new direction; the *why* (hypothesis: the dip is where the profile's curvature, and therefore near-unloading moments, are most active) is flagged but **not root-caused**, matching this same file's own established precedent from issue #228 for not chasing a mechanism outside a pass's scope.

## Acceptance criteria satisfied (issue #250)

1. ✅ Reject/hold the accelerometer update below a derived threshold — done, unconditionally, independent of the existing opt-in trust band.
2. ⬜ Left out — see below.
3. ⬜ Left out — see below.
4. ✅ Test that drives specific force to zero and asserts the estimator does not invert — two tests, covering both the first-sample and sustained cases.

## Deliberately left out

- **Asks 2 and 3** (defined recovery behaviour for a sustained low-g drop where gyro-only integration drifts; whether this deserves a `Faults` bit). Both are real design questions the issue itself only asks to "decide" or "consider" — no stated numeric target — and are exactly the kind of judgement call this session's scope discipline says to flag rather than invent an answer for unattended. Left for a deliberate pass.
- **Whether the rolling-vs-steady headline is true at some other grade or parametrisation, independent of this bug.** Not re-examined — the original "transitions cost margin" hypothesis is neither confirmed nor refuted here, only shown to have been entangled with a defect at the specific 10% point it was pinned at.
- **Root-causing why the dip specifically inverts.**
- **ADR-0011 itself and issue #227 are not edited or closed.** This finding bears directly on both — #227 is literally about whether the freeze-and-pin generalises past flat ground, and this session found a second, independent reason the terrain-scenario numbers around it were unreliable — flagged here for the COO, not resolved. `docs/` is COO turf regardless.
- **`control-ffi`'s own default `Params`/FFI field defaults left untouched** — the new floor already protects every FFI consumer unconditionally, so no default needed to change there.

## Verification

- `cargo test -p control-core` — 37/37 (35 pre-existing + 2 new).
- `cargo test --workspace` — clean, no regressions.
- `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`, `cargo run -p xtask -- gate` — all clean.
- `python3 -m pytest tests/` — 333 passed, 5 xfailed (same counts as the pre-fix baseline; the 4 that flipped are re-pinned to measured numbers, not skipped or deleted).
- `python3 .github/policy_check.py` — all hard checks pass.

Closes #250.

---
_Generated by [Claude Code](https://claude.ai/code/session_019cu9P4DJfqwvnZCf2rw9Cj)_

### DIFF
--- a/crates/control-core/src/lib.rs
+++ b/crates/control-core/src/lib.rs
@@ -83,6 +83,38 @@ pub trait Estimator {
 /// This is the standard trick and it is cheap, but it is not free: while
 /// coasting there is no attitude reference at all, so the band must be wide
 /// enough that ordinary riding does not trip it. Zero disables the gate.
+///
+/// # A near-zero specific force has no direction at all (issue #250)
+///
+/// The band above is opt-in and centred on deviation *from g* — a caller can
+/// legitimately leave it at zero. This next check is neither: it is
+/// unconditional, and it is not about deviation from g, it is about the raw
+/// magnitude approaching zero.
+///
+/// `atan2` never refuses. Fed a vector of (near) zero length it still returns
+/// a confident angle — `atan2(0, -0) = 180°` exactly, "upside down" — but
+/// that angle carries no information, because a vanishing vector has no
+/// defined direction. Trusting it is not a tuning problem, it is a category
+/// error: there is nothing to fuse in. A real board reads this on any brief
+/// unloading — cresting a rise, a kerb, a drop, a jump — all ordinary events
+/// on the terrain this board is meant for, not exotic edge cases.
+///
+/// [`MIN_TRUSTED_ACCEL_MAG_M_S2`] is derived, not picked, from the same
+/// envelope [`ComplementaryFilter::with_trust_band`]'s doc reasons about:
+/// swept over the outer loop's own ±0.86 m/s² commanded-acceleration limit
+/// and every pitch up to the ±20° fallen threshold, specific-force magnitude
+/// never moves more than ~0.33 m/s² away from g in either direction — and
+/// every disturbance this repo has characterised (the 400 N·s nominal kick,
+/// the 260–320 N·s envelope sweep) is a horizontal shove, which by
+/// `‖(g·sinθ+a, g·cosθ)‖ ≥ g` for any horizontal `a` can only ever raise the
+/// magnitude, never lower it. So nothing this codebase has measured, and
+/// nothing the controller itself can command, drives magnitude down at all —
+/// only genuine unloading does that, running the magnitude toward zero as
+/// support vanishes. Half of g leaves wide separation from both sides: far
+/// above zero (so it never mistakes a real low-g moment for noise) and far
+/// below the ~9.5 m/s² floor of grounded, in-envelope operation.
+pub const MIN_TRUSTED_ACCEL_MAG_M_S2: f32 = 9.81 / 2.0;
+
 #[derive(Debug, Clone, Copy)]
 pub struct ComplementaryFilter {
     tau_s: f32,
@@ -122,11 +154,18 @@ impl ComplementaryFilter {
     /// Is this sample's specific force consistent with gravity plus the
     /// acceleration we already know about?
     fn accel_trusted(&self, accel: [f32; 3], forward_accel_m_s2: f32) -> bool {
+        let (x, y, z) = (accel[0] - forward_accel_m_s2, accel[1], accel[2]);
+        let mag = libm::sqrtf(x * x + y * y + z * z);
+
+        // Unconditional (issue #250): a vector this small has no direction to
+        // trust, regardless of whether the caller configured a trust band.
+        if mag < MIN_TRUSTED_ACCEL_MAG_M_S2 {
+            return false;
+        }
+
         if self.accel_trust_band_m_s2 <= 0.0 {
             return true;
         }
-        let (x, y, z) = (accel[0] - forward_accel_m_s2, accel[1], accel[2]);
-        let mag = libm::sqrtf(x * x + y * y + z * z);
         libm::fabsf(mag - 9.81) <= self.accel_trust_band_m_s2
     }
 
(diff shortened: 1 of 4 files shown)


MikePaNtZ:
Two CI notes, both resolved, one left intentionally red:

1. **`policy` failed** — I'd pushed a one-line fix to `scripts/render_scenario.py`'s `--terrain-compare-grade` default (same stale 10% assumption this PR already re-derives in `tests/test_terrain.py`), but that file is CODEOWNERS-owned by Digital Content Production, not Senior Controls. Reverted the out-of-turf commit rather than override it. `policy` should now pass.
2. **`publish-sim-artifact` will stay red** as a result — it renders the terrain comparison clip at the same stale 10% default, which no longer diverges after this PR's estimator fix. This is not a required check, and the fix is out of my turf, so I filed #265 for Digital Content Production instead of touching it here.

Everything required (`rust`, `sim`, `policy`) should be green on the latest commit.

---

MikePaNtZ:
Queued for auto-merge. The one red check is `publish-sim-artifact`, which is **not** in master's required set — protection requires `rust`, `sim`, `policy` only. Per the repo rules that job is deliberately not required, because it runs on push to master and requiring it would hang every PR waiting for a check that never reports. So `UNSTABLE` here is expected and not a blocker.

Issue #286, opened 2026-08-16
The image cannot do wifi at all -- wpasupplicant is missing, so the staged NetworkManager profiles are inert


MikePaNtZ:
Found on hardware, 2026-08-16.

## What happens

`flash_pi.sh` stages three NetworkManager wifi profiles onto the boot partition, `firstboot_install.sh` installs them, and they are correct — right SSID, `autoconnect=true`, priority 100. They can never work, because the image has no wifi backend:

```
$ which wpa_supplicant
(nothing)

$ nmcli d
DEVICE  TYPE      STATE         CONNECTION
wlan0   wifi      unavailable   --

$ sudo nmcli d wifi connect MMMKWiffy
Error: No network with SSID 'MMMKWiffy' found.
```

`rfkill` is clear on `phy0`, soft and hard, and `brcmfmac` loads normally at boot. The radio is fine. NetworkManager reports `unavailable` because it has nothing to drive it.

## Why it matters

This is what makes #285 unrecoverable rather than merely annoying. A headless card with a full disk and no network has no way in — no wifi, no package install, and (without #284's fix) no trustworthy SSH either. The credential staging in `flash_pi.sh` implies the image is ready to be headless. It is not.

## Evidence

`pins.env` at `2e41fc7`:

```
PINNED_PACKAGES="
rt-tests=2.6-1.1
can-utils=2023.03-1+b2
"
```

## Fix

Pin `wpasupplicant` in `PINNED_PACKAGES`. Note the Debian package name is **`wpasupplicant`**, one word — `wpa_supplicant` does not resolve.

Two more while that file is open, both hit during this bring-up:

- **`python3-spidev`** is not pinned. Candidate in trixie is `3.6-1+b6`. Needed for any SPI encoder work, which is the next hardware step.
- **`cpufrequtils` does not exist in trixie.** Anything that references it will fail. The governor can be set through sysfs with no package, and made persistent with a small systemd unit; `linux-cpupower` is the packaged equivalent if one is wanted.

AC-1 requires every directly-installed package to be pinned, so these belong in `pins.env` rather than in a runbook step.

## Acceptance

- A freshly flashed card associates with a staged wifi profile on first boot, with no console intervention.
- `pins.env` carries `wpasupplicant` and `python3-spidev` at explicit versions.
- No reference to `cpufrequtils` survives anywhere in the tree.

## Workaround applied to the current card

[...]

Issue #285, opened 2026-08-16
First boot never expands the rootfs -- the image arrives 100% full and nothing works


MikePaNtZ:
Found on hardware, 2026-08-16, flashing the Stage-0B image to a 32 GB card.

## What happens

The image ships a fixed 651 MB partition 2 and never grows it. On a 32 GB card:

```
/dev/mmcblk0p2   651M  634M     0  100% /
/dev/mmcblk0p1    49M   48M  400K  100% /boot/firmware
```

Zero bytes free on `/`. The visible symptom is a boot that reaches a login prompt and then fails at everything:

```
systemd-journald[278]: Failed to open user journal file, falling back to
system journal: No space left on device
```

`apt update` cannot write a package index, so every `apt install` returns *"Unable to locate package"* — which reads as a missing package name and sends you looking in the wrong place entirely.

## Why it is worse than it sounds

The image also ships **no partition tool** — no `parted`, no `growpart`, no `sfdisk`. So the operator cannot fix it on the Pi with what is on the Pi, and cannot install a tool either, because the disk is full and there is no network (see the companion issue on `wpasupplicant`). On a headless card with no serial adapter this combination is unrecoverable.

## Evidence it is not already handled

```
$ git grep -i -e resize2fs -e growpart -e "expand.*root" 2e41fc7 -- scripts/
(no match)
```

## Fix

Expand on first boot, in `firstboot_install.sh`, before anything else runs. `resize2fs` **is** present in the image, so only the partition step is missing. Either:

- ship `parted` or `cloud-guest-utils` and call `growpart` + `resize2fs`, or
- grow partition 2 to the end of the device with `sfdisk`, then `resize2fs` after the next boot.

## Acceptance

- A freshly flashed card reports `/` sized to the card, not 651 MB, after first boot.
- The build fails, or the first boot logs a loud warning, if expansion did not happen.
- A partition tool is pinned in `pins.env` so the recovery path exists on the card.

## Workaround applied to the current card

Patched the MBR sector count for partition 2 by hand, then:

```sh
sudo reboot && sudo resize2fs /dev/mmcblk0p2
```

Result: `30G  830M  28G  3% /`.

Relates to #182. Companion to #284.

Issue #265, opened 2026-08-08
scripts/render_scenario.py's --terrain-compare-grade default (10%) is stale after issue #250's estimator fix


MikePaNtZ:
Owner: Digital Content Production (`scripts/render_scenario.py` is CODEOWNERS:148, not Senior Controls turf — found this out the hard way, see below).

## What

`scripts/render_scenario.py --scenario terrain --compare` publishes the "truth pitch vs attitude estimate" comparison clip that `publish-sim-artifact` uploads on every green `master` push. It defaults `--terrain-compare-grade` to `10.0` and asserts truth completes the ride (raising if not) — the assumption being that at 10% peak grade, truth pitch finishes the rolling-terrain ride and the estimate does not, so the two panes visibly diverge.

Issue #250 (PR #264) fixed a real estimator defect (a near-zero specific-force reading resolving through `atan2(0, -0)` to a confident, arbitrary ~180°). One measured side effect: the terrain scenario's divergence-at-10% no longer holds — **both** truth and estimate now complete a 10% ride, so a render at the current default would publish a "comparison" clip whose two panes reach the *same* outcome, contradicting its own caption.

`tests/test_terrain.py` (Senior Controls turf) has already been re-derived to the newly measured divergence window: **10.6–11.4%** peak grade, with **11.0%** picked as a stable midpoint (11.6% already flips back — a non-monotonic-near-boundary signature this codebase has seen before in issue #24 AC2's disturbance sweep). See `tests/test_terrain.py::test_the_estimator_costs_terrain_envelope` on PR #264 for the exact sweep data.

## Why I'm not just fixing this myself

I (Senior Controls, this session) initially pushed the one-line fix directly to `scripts/render_scenario.py` on PR #264, since it's the same numeric assumption I'd already re-derived elsewhere. CI's `policy` gate caught it: that file is CODEOWNERS-owned by Digital Content Production, not Senior Controls, and doc-drift also flagged that `docs/sim-impulse-response.md` and `docs/web-artifact-pipeline.md` both describe this script and would need a matching update. I reverted the out-of-turf commit rather than override it — this issue is the correct handoff instead.

## Ask

[...]

PR #287, opened 2026-08-17
fix(controls): strip baked-in SSH host keys from the Stage-0B image (#284)

### LINKED ISSUE #284: The Stage-0B image ships SSH host keys generated in the CI build container
Found while bringing up the RT image on hardware, 2026-08-16.

## What is wrong

The image carries pre-generated SSH host keys. The key on the first flashed card was:

```
256 SHA256:fCKX0hTFcCyJcVufQmvgZuZRUwi732SA9tMO15RgmuE root@01e4d9e392f9
```

`01e4d9e392f9` is a Docker container ID. The key was made inside the `debian:trixie` build container and baked into the rootfs.

## Why it matters

1. **Every card flashed from this image has the same host key.**
2. The image is published as a **public CI artifact** from a **public** repo, so the private half of that key is downloadable by anyone.
3. SSH host authentication for every Overboard Pi is therefore worth nothing. Anyone on the same network can impersonate the Pi and the operator's client will not warn.

This also produces a confusing false positive in normal use: two cards from the same image present one key, and a re-flash presents the *same* key, so the usual "REMOTE HOST IDENTIFICATION HAS CHANGED" signal no longer means what an operator assumes.

## Evidence it is not already handled

```
$ git grep -l -i "ssh_host\|ssh-keygen -A" 2e41fc7 -- scripts/ .github/
(no match)
```

## Fix

Delete `/etc/ssh/ssh_host_*` at image build time, and regenerate on first boot. `firstboot_install.sh` is the natural place:

```sh
rm -f /etc/ssh/ssh_host_*
ssh-keygen -A
systemctl restart ssh
```

Debian also ships `regenerate-ssh-host-keys.service` for exactly this.

## Acceptance

- The built image contains **no** `/etc/ssh/ssh_host_*`.
- Two cards flashed from one image present **different** host key fingerprints.
- A CI assertion fails the build if any `ssh_host_*` file is present in the rootfs.

## Workaround applied to the current card

```sh
sudo rm -f /etc/ssh/ssh_host_* && sudo ssh-keygen -A && sudo systemctl restart ssh
```

Fingerprint went from `root@01e4d9e392f9` to `root@overboard`.

Relates to #182.

MikePaNtZ:
## What changed

Issue #284, found on hardware 2026-08-16: `openssh-server`'s postinst generates SSH host keys at **package-install time**, inside the CI build container — so every card built from a given container state ships the identical private key, published as a public CI artefact anyone can download and use to impersonate the board.

- **`scripts/pi/image/layer/overboard-base.yaml`**: new `customize-hooks` entry strips `/etc/ssh/ssh_host_*` from the built rootfs, with a self-check that fails the hook itself if any key survives the `rm`.
- **`scripts/pi/image/layer/overboard-base.rootfs-overlay/etc/systemd/system/overboard-ssh-hostkeys.service`**: new first-boot unit, `Before=ssh.service`, that runs `ssh-keygen -A` to generate a fresh, unique pair on the card. `ssh-keygen -A` only creates key types that don't already exist, so this is idempotent and needs none of `overboard-firstboot.service`'s marker-file gating — there's no staged, single-use credential to shred here.
- **`scripts/pi/build_image.sh`**: a second-line-of-defence CI assertion (same style as the existing kernel-selection check) that fails the build if any `ssh_host_*` file is present anywhere in the built rootfs.

## Acceptance criteria (from #284)

- [x] The built image contains no `/etc/ssh/ssh_host_*` — the customize-hook removes them, and self-verifies.
- [x] Two cards flashed from one image present different host key fingerprints — `overboard-ssh-hostkeys.service` generates a fresh pair per card, per boot-until-present.
- [x] A CI assertion fails the build if any `ssh_host_*` file is present in the rootfs — `build_image.sh`'s new section.

## Verification

- `python3 scripts/pi/check_layer_headers.py`, `python3 scripts/pi/check_no_secrets.py`, `python3 scripts/pi/check_image_pin.py` — all clean.
- `python3 .github/policy_check.py` — all hard checks pass (pre-existing advisories only, none touched by this change).
- `bash -n scripts/pi/build_image.sh` — clean; the new hook's shell fragment parses standalone.
- Parsed `overboard-base.yaml`'s `customize-hooks` list with PyYAML after stripping the `METABEGIN`/`METAEND` header, confirmed 7 hooks including the new one, in order.
- **Could not verify end-to-end**: this sandbox has no privileged Docker / arm64 host to actually run `build_image.sh`'s 90-minute `rpi-image-gen` build, and has no `ssh-keygen` binary installed at all to smoke-test the unit's `ExecStart` locally. `/usr/bin/ssh-keygen` is the standard path shipped by Debian's `openssh-client` (a dependency of `openssh-server`, already pinned into this image via the `openssh-server` package in `overboard-base.yaml`) — verified the path exists as a matter of Debian packaging convention, not by running it. The `pi-image` workflow's `build-image` job will run the new build-time hook and CI assertion for real on this PR.

## Deliberately left out

- The `Before=ssh.service` ordering relies on `ssh.service` also being pulled into the `multi-user.target` transaction (the same pattern `overboard-firstboot.service` already uses successfully for `NetworkManager.service`) rather than an explicit `DefaultDependencies=no` / `sysinit.target` ordering — kept to the codebase's existing, proven unit style rather than inventing new low-level ordering I have no way to boot-test here.
- Did not touch `pins.env` or `PINNED_PACKAGES` — this issue is about key generation timing, not package pins.
- Companion issues #285 (rootfs never expands on first boot) and #286 (wifi backend missing, `wpasupplicant` unpinned) are separate, each its own PR — not bundled in here, though all three were found together on the same hardware bring-up session.

Closes #284.

---
_Generated by [Claude Code](https://claude.ai/code/session_0153q8XRrzUbBWsE8XJfYMRT)_

### DIFF
--- a/scripts/pi/build_image.sh
+++ b/scripts/pi/build_image.sh
@@ -100,6 +100,25 @@ else
   exit 1
 fi
 
+section "no baked-in ssh host keys (#284)"
+
+# openssh-server's postinst generates host keys at package-install time,
+# inside THIS build container -- every image built from a given container
+# state would otherwise ship the identical private key, published as a
+# public CI artefact anyone can download and use to impersonate the board.
+# overboard-base.yaml's customize-hooks strip them from the rootfs; this is
+# the second line of defence in case that hook is ever weakened without
+# anyone noticing until a card ships.
+leaked_keys="$(find "$WORK/work" -path '*/etc/ssh/ssh_host_*' 2>/dev/null)"
+if [ -n "$leaked_keys" ]; then
+  echo "FATAL: /etc/ssh/ssh_host_* found in the built rootfs -- every card"
+  echo "       flashed from this image would share the same SSH identity."
+  echo "$leaked_keys" | sed 's/^/         /'
+  exit 1
+else
+  echo "OK: no ssh_host_* files in the built rootfs"
+fi
+
 section "collect artefacts"
 mkdir -p "$OUT"
 rm -f "$OUT"/*
--- a/scripts/pi/image/layer/overboard-base.rootfs-overlay/etc/systemd/system/overboard-ssh-hostkeys.service
+++ b/scripts/pi/image/layer/overboard-base.rootfs-overlay/etc/systemd/system/overboard-ssh-hostkeys.service
@@ -1,0 +1,24 @@
+[Unit]
+Description=Overboard SSH host key generation
+# openssh-server's postinst generates host keys at PACKAGE-INSTALL time --
+# inside the build container, not on the card. Every image built from that
+# container would otherwise ship the identical private key, published as a
+# public CI artefact anyone can download and use to impersonate the board
+# (#284). overboard-base.yaml's customize-hooks strip /etc/ssh/ssh_host_*
+# from the built rootfs; this unit regenerates a fresh, unique pair before
+# sshd can start, which needs to exist for ssh.service to come up at all.
+#
+# `ssh-keygen -A` only creates key types that do not already exist, so this
+# is idempotent and safe to leave enabled on every boot -- unlike
+# overboard-firstboot.service there is no staged, single-use credential here
+# to shred, so no marker-file gating is needed.
+Before=ssh.service
+Wants=ssh.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/bin/ssh-keygen -A
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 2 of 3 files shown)


MikePaNtZ:
`publish-sim-artifact` is red on this PR (`Verify the terrain clips were actually produced` failing with `the estimate completed the ride too, so this is not the comparison it is captioned as -- re-derive the envelope with Senior Controls`), but it's pre-existing on master itself: the latest master push (`d23f055`, this PR's own base) fails at the identical step for the identical reason. Ties to #265 (stale `--terrain-compare-grade` default after #250's estimator fix), which is Digital Content Production's turf, not touched by this PR. It's also not a required check (`rust` and `sim` are). Not fixing here — will re-run when the base recovers.

---
_Generated by [Claude Code](https://claude.ai/code/session_0153q8XRrzUbBWsE8XJfYMRT)_

---

PR #288, opened 2026-08-17
fix(controls): expand the root filesystem on first boot (#285)

### LINKED ISSUE #285: First boot never expands the rootfs -- the image arrives 100% full and nothing works

MikePaNtZ:
## What changed

The Stage-0B image ships a fixed ~651 MB partition 2, sized to its build-time content by rpi-image-gen's `image-rpios` layer, with no knowledge of the card it gets flashed to. A freshly flashed card boots 100% full — no room for the journal, `apt`, or anything else — and the image ships no partition tool to recover from the Pi itself (found on hardware, 2026-08-16).

Adds `overboard-rootfs-expand.service`, a first-boot `systemd` oneshot ordered **before** `overboard-firstboot.service`, whose script:
1. Resolves the root partition/disk via `findmnt` + `lsblk` + sysfs (works for `mmcblk0pN` or `sdaN`, not hardcoded).
2. Grows the partition with `growpart` (from `cloud-guest-utils`).
3. Grows the ext4 filesystem with `resize2fs`.

Idempotent by asking each tool its own "nothing to do" state (`growpart`'s `NOCHANGE`, `resize2fs`'s "already ... blocks long") rather than a completion marker — a marker written before the job is truly done is exactly the bug `overboard-firstboot.service` was rewritten to avoid, and it would be worse here: a card permanently stuck at 651 MB rather than just missing credentials for one boot.

**`parted` was tried first and rejected.** A loop-device rehearsal of this exact scenario (grow the currently-mounted root partition, live) showed `parted -s -f resizepart` refuses outright on "Partition ... is being used", even with `--fix`. `growpart` — the tool cloud-init images use for the identical case — succeeded in the same rehearsal.

## Acceptance criteria (from #285)

- ✅ A freshly flashed card reports `/` sized to the card, not 651 MB, after first boot — verified as far as this sandbox allows (see below); real-hardware confirmation is still open.
- ✅ The first boot logs a loud warning if expansion did not complete: `growpart` failures that aren't `NOCHANGE` are logged as `ERROR` and exit 1; a `resize2fs` that can't complete online is logged as `WARNING` and retried automatically next boot (never silently skipped, never wedges the boot).
- ✅ A partition tool is pinned in `pins.env`: `cloud-guest-utils` (growpart) and `fdisk` (growpart's `sfdisk` backend for this image's MBR layout).

## Verified locally

- `shellcheck` clean, `bash -n`, YAML parses, `policy_check.py`, `check_layer_headers.py`, `check_image_pin.py`, `check_no_secrets.py` all green.
- **Loop-device rehearsal**: built a small ext4 filesystem on an undersized, mounted partition and ran the actual script logic against it. `growpart` correctly resized the mounted partition (27 MB → 170 MB in the test) and correctly reported `NOCHANGE` on a second run (idempotency confirmed). `resize2fs`'s online-resize step hit "Permission denied" in this sandboxed container — consistent with a nested-container ioctl restriction on loop-mounted filesystems, not a script defect, since online ext4 resize of a mounted filesystem after partition growth is the exact, decade-proven mechanism cloud-init images use on every VM boot. **Could not verify the resize2fs step end-to-end on real hardware** — the script's warn-and-retry-next-boot path exists specifically to be safe if that step doesn't complete on the first attempt.

## Deliberately left out

- **#286** (wpasupplicant/wifi missing) and **#284** (SSH host keys baked from the build container, already has an open PR) are companion issues found in the same hardware bring-up — each is its own increment, not bundled here.
- **The two new package pins could not be checked against the primary source.** `packages.debian.org` and `deb.debian.org` are both blocked by this session's network egress policy (confirmed via the proxy status endpoint — `deb.debian.org` shows a policy-denied 403). The pinned versions (`cloud-guest-utils=0.33-1`, `fdisk=2.41-5`) are a best-effort read from web search results, cross-checked against what Ubuntu noble's own reachable archive resolves today (`cloud-guest-utils` matched exactly at 0.33-1; `fdisk` did not, since noble's util-linux generation predates trixie's and was never expected to match) — flagged as unconfirmed in a `pins.env` comment rather than asserted as verified. `verify-pins` (not a required check, but it does reach the live archive) will confirm or correct them on this PR.

Closes #285

---
_Generated by [Claude Code](https://claude.ai/code/session_01E6iEJppMjEKvp6jtsykcvS)_

### DIFF
--- a/scripts/pi/image/layer/overboard-base.rootfs-overlay/etc/systemd/system/overboard-rootfs-expand.service
+++ b/scripts/pi/image/layer/overboard-base.rootfs-overlay/etc/systemd/system/overboard-rootfs-expand.service
@@ -1,0 +1,29 @@
+# Grows the root partition and filesystem to fill the card, at boot.
+#
+# rpi-image-gen's image-rpios layer sizes partition 2 to its build-time
+# content (~651 MB today) and never grows it -- a fresh flash boots 100%
+# full and is unrecoverable headless (#285). Ordered before
+# overboard-firstboot so credential installation, journald, and everything
+# after it has room to work.
+[Unit]
+Description=Overboard root filesystem expansion
+After=local-fs.target
+Before=overboard-firstboot.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/local/sbin/overboard-rootfs-expand
+
+# Generous but bounded, matching overboard-firstboot.service's own reasoning:
+# Type=oneshot disables TimeoutStartSec by default, so an unbounded resize
+# could wedge the boot with no rescue. `resize2fs` on a card this size is a
+# matter of seconds, not minutes.
+TimeoutStartSec=90
+
+# The journal is the only record on a headless board with no console user.
+StandardOutput=journal
+StandardError=journal
+
+[Install]
+WantedBy=multi-user.target
--- a/scripts/pi/image/layer/overboard-base.rootfs-overlay/usr/local/sbin/overboard-rootfs-expand
+++ b/scripts/pi/image/layer/overboard-base.rootfs-overlay/usr/local/sbin/overboard-rootfs-expand
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+# Runs ON THE PI, at boot, before overboard-firstboot. Shipped inside the
+# image (I1).
+#
+# rpi-image-gen's image-rpios layer sizes partition 2 to its BUILD-TIME
+# content -- currently ~651 MB -- with no knowledge of the card the image
+# will end up on. Left alone, a freshly flashed card boots 100% full: no
+# free space for the journal, for apt, or for anything else (#285), and the
+# image ships no partition tool to fix it from the Pi itself.
+#
+# growpart, NOT parted: verified in a loop-device rehearsal of this exact
+# scenario (grow the currently-mounted root partition live) that parted
+# refuses outright -- `parted -s -f resizepart` exits 1 on "Partition ...
+# is being used" even with --fix, because that prompt is a hard abort in
+# script mode, not one --fix answers. growpart is what cloud-init images
+# use for precisely this case (grow the mounted root partition at boot, no
+# reboot) and it succeeded in the same rehearsal. sfdisk is growpart's MBR
+# backend (this image's layout, per image/config/overboard.yaml) -- ships
+# in the `fdisk` package, not `cloud-guest-utils` itself.
+#
+# No completion marker: growpart and resize2fs each report their own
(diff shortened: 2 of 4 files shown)


MikePaNtZ:
`publish-sim-artifact` is red on this PR, but it's a pre-existing failure — `master`'s current HEAD (`d23f055`, pushed before this PR existed) fails the same way, on the same step ("Verify the terrain clips were actually produced" / the terrain-estimate-vs-truth divergence assertion in `ci.yml`). Unrelated to this PR's diff (Pi image only). Not a required check either way. Will re-run once `master` recovers.

`verify-pins` is also red — see the comment in `scripts/pi/pins.env` next to the `fdisk` pin for what's known so far (the failure signature doesn't match a plain wrong-version error, and this session cannot reach the Debian archive to dig further). Also not required.

---

Issue #230, opened 2026-08-05
AC-5 cannot be measured on the current thermal solution — the Pi throttled in 4 minutes of load, and AC-5 asks for 30


MikePaNtZ:
Measured on the CEO's Pi 5 Rev 1.1, 2026-08-05, first load test (`runbook-pi-first-boot.md` §6/§7 + `stress-ng --cpu 4 --sock 2`).

```
throttled=0xe0000
temp=80.7'C
```

## Decoding

| Bit | Meaning | Set? |
|---|---|---|
| 16 | under-voltage **has occurred** | **no** ✅ |
| 17 | arm frequency capping **has occurred** | **yes** ⚠️ |
| 18 | throttling **has occurred** | **yes** ⚠️ |
| 19 | soft temperature limit **has occurred** | **yes** ⚠️ |

Low nibble clear, so nothing was *still* active when read — these are the sticky "has occurred since boot" flags.

**The good news, and it closes a parts-gate item:** bit 16 is clear. The PSU carried four stressed cores plus network workers with **no under-voltage dip**, so the supply is adequate. That had been an assumption.

**The problem:** the board reached **80.7 °C** and hit the soft temperature limit, capped ARM frequency, and throttled — after roughly **4 minutes 20 seconds** of load.

## Why this blocks AC-5 rather than merely annoying us

AC-5 is **`cyclictest -m -Sp95 -i 2000 -D 30m` under `stress-ng` CPU + network load**. Thirty minutes. This board throttled in under five.

A throttled CPU is not a slower CPU held constant — it is a CPU whose clock is *changing underneath the measurement*, which is a jitter source in its own right. So a 30-minute AC-5 run on this thermal solution does not produce a pessimistic-but-usable number; it produces an **uninterpretable** one, in exactly the way the parent design warns a bad CAN transport would "poison the go/no-go number." The same argument applies here, and it applies to the AC-5 gate that precedes AC-6.

The Stage-0B runbook already asserted this without evidence — "a throttling Pi is a jitter source, and jitter is the entire deliverable of Stage 0B." It now has evidence.

## Contaminated data, recorded so it is not reused

The loaded profiler run was taken **during** this throttling, so it is not a clean baseline. Reporting it anyway, flagged:

wake latency, `SCHED_FIFO:80`, `6.18.34+rpt-rpi-2712`, µs:

| | p50 | p99 | p99.9 | max |
|---|---|---|---|---|
[...]

PR #291, opened 2026-08-21
docs(controls): issue #168's Pacer fix and jitter-percentiles ask are already closed

### LINKED ISSUE #168: The 500 Hz host is bursty, not late: ~15 ms stall then an 8-tick catch-up burst, and the Pacer comment says the opposite
Filed by the COO from independent verification of #161 / PR #167. **Not a launch blocker** — the game absorbs this — but it is mis-described in the code today, and the mis-description is the dangerous part.

## What the numbers actually say

Measured twice, second time on a deliberately idle machine to rule out load from concurrent agent sessions:

```
inter-packet interval (ms): mean=1.9723  p50=0.0921  p99=15.0325  max=17.6350
measured tick rate (mean): 506.81 Hz  (5069 packets, 0 parse failures)
missed deadlines: 4528 / 6479 ticks (~70%)
```

Mean 2 ms, **median 0.09 ms**, p99 15 ms. That is not "most ticks are slightly late." The loop **stalls ~15–17 ms, then fires roughly eight ticks back-to-back** to catch up. It reproduces on an idle machine, so it is not contention from other work on the box.

## Probable cause

macOS **timer coalescing**. Short `thread::sleep` durations get quantized toward the ~16 ms boundary unless the thread asks for a real-time policy, so a 2 ms sleep becomes a ~16 ms sleep. `Pacer` then correctly returns `Duration::ZERO` until it catches back up, producing the burst. The ~16 ms period and the 8-tick burst size are both consistent with this.

This is an OS scheduling artifact, **not a defect in the control law or the pacing arithmetic**. The aggregate rate is right precisely because `Pacer` uses absolute deadlines and does not accumulate drift. It is also a good argument that the ~70% figure says more about stock macOS than about this code — the PREEMPT_RT Linux target is where the deadline bar is meant to be met, and that remains untested.

## The actual defect: the comment claims the opposite of what happens

`crates/sim-host/src/pacer.rs` says advancing by one period means:

> a single slow cycle costs one miss, not a cascading pile-up of catch-up iterations trying to make up lost ground

**A pile-up of catch-up iterations is exactly what the data shows.** After a 15 ms stall the deadline is ~7 periods in the past, so the loop runs ~7 zero-sleep iterations back-to-back. The behaviour is defensible — sim time stays locked to wall time on average, which is what a fixed-step sim needs — but the comment tells the next reader the opposite, and it is the kind of thing someone will rely on at 2 a.m.

## Asks

1. **Fix the comment** to describe the real behaviour: bounded catch-up bursts, correct on average, deliberately chosen over resetting the deadline to `now` (which would silently drop sim time).
2. **Consider a real-time thread policy on macOS** (`thread_policy_set` / `THREAD_TIME_CONSTRAINT_POLICY`, or `mach_wait_until`) so the host is smooth on the dev machines everyone actually demos from. Tuesday work, not weekend work.
3. **Report jitter percentiles, not just a miss count.** "70% missed" and "p99 = 15 ms with a correct mean" are the same fact, but the first reads as catastrophic and the second is actionable. A raw miss count against a zero-slack deadline is a misleading headline number on any non-realtime OS.
4. **The miss count travels out-of-band in `/tmp/overboard-sim-host-stats.txt`.** Fine for W1; it should become a wire field on the next version bump rather than staying a file.

## Consequence already handled

The Game Engineer has been told **not** to reduce `RenderDelaySeconds` below ~0.05 s — the client buffer has to cover a 17.6 ms worst-case gap, and they were otherwise planning to tune it down. Their interpolate-never-extrapolate design absorbs the bursts cleanly.

MikePaNtZ:
## What changed and why

Docs-only, no code changes. Adds
`roles/senior-controls/log/2026-08-21-issue-168-verify-stale.md` recording that issue #168 is
stale-closed, not stale-open.

## What #168 asked for

Four asks: (1) fix `Pacer`'s inverted doc comment about catch-up bursts, (2) consider a
real-time thread policy on macOS, (3) report jitter percentiles instead of a bare miss count,
(4) promote the miss count to a wire field. The issue's own text defers asks 2 and 4 ("Tuesday
work, not weekend work" / "fine for W1").

## What I verified, not assumed

- Ask 1: `crates/sim-host/src/pacer.rs`'s `wait_for_next` doc comment already carries a
  `CORRECTION (issue #168)` block describing the real bounded-catch-up-burst behaviour.
- Ask 3: `Pacer::jitter_percentiles()` (nearest-rank p50/p99/max over a bounded window) is
  implemented, tested (4 dedicated tests), and wired into `host.rs`'s stats file and
  `wire-probe`'s report line alongside the existing miss count.
- Both landed in `120c3dc` (PR #267, 2026-08-08). That PR's own log entry
  (`roles/senior-controls/log/2026-08-08-issue-168-jitter-percentiles.md`) states up front it
  "references #168, does not close it" — and no later PR added a `Closes #168`, so the issue sat
  open with nothing left to do against it. Same shape as #230's stale-closure
  (`roles/senior-controls/log/2026-08-09-issue-230-verify-stale.md`).
- Re-ran `cargo test -p sim-host --lib pacer` fresh on this checkout: 9/9 passing, unchanged.
- `python3 .github/policy_check.py`: all hard checks pass (advisory-only warnings, pre-existing).

## Acceptance criteria satisfied

Issue #168's two in-scope asks (1 and 3) are landed and verified above. Asks 2 and 4 are the
issue's own explicit deferrals, not unmet acceptance criteria.

## Deliberately left out

- Ask 2 (RT thread policy on macOS) and ask 4 (wire field for the miss count) — both are the
  issue's own stated deferrals. Either is a fine future increment on its own acceptance
  criterion, not something this PR should invent or bundle in.
- No code changes anywhere — this is a record-keeping fix for a decision that already shipped.

## Also found, out of scope (see log entry for detail)

- Issue #132's own default action needs a bench-fitted `kt` from Sr. Mechanical & Systems first;
  `sim/scenarios/plant.py`'s `KT_NM_PER_A` is still the documented unfitted 0.7 placeholder, so
  #132 isn't yet actionable by Senior Controls.
- Issue #182's remaining increments (I6 platform-contract manifest, I7 sim-in-the-loop
  self-test, I8 motor-in-the-loop) are each real architecture work past a single unattended
  increment, several needing hardware in hand.

Closes #168.

---
_Generated by [Claude Code](https://claude.ai/code/session_016mfUL15T2pAnPTnkx7jaUb)_

### DIFF
--- a/roles/senior-controls/log/2026-08-21-issue-168-verify-stale.md
+++ b/roles/senior-controls/log/2026-08-21-issue-168-verify-stale.md
@@ -1,0 +1,73 @@
+# 2026-08-21 — Issue #168: the Pacer bug and jitter-percentiles ask are already closed
+
+Cron dispatch pass. Open issues labeled `role:senior-controls`: #182 (Stage-0B umbrella, active,
+increments I6–I8 are hardware/architecture work well past a single scoped increment), #168, #132
+(blocked — its own default action requires Sr. Mechanical's bench-fitted `kt` first;
+`sim/scenarios/plant.py`'s `KT_NM_PER_A` is still the unfitted 0.7 placeholder, so the
+precondition for "Senior Controls owns the retune" does not exist yet), #61 (`Dispatch: COO
+only`, reserved). Of that set, #168 had no PR open against it and turned out to already be
+resolved.
+
+## What #168 asked for
+
+Filed after the COO's independent verification of the 500 Hz host's pacing: `Pacer`'s own doc
+comment claimed a single slow cycle costs one miss, "not a cascading pile-up of catch-up
+iterations" — measured data (macOS timer coalescing, ~15 ms stalls followed by ~8-tick bursts)
+showed exactly that pile-up, so the comment described the opposite of the real behaviour. Four
+asks: (1) fix the comment, (2) consider a real-time thread policy on macOS, (3) report jitter
+percentiles instead of a bare miss count, (4) promote the miss count to a wire field. The issue
+states its own deferrals in its own text — ask 2 is "Tuesday work, not weekend work," ask 4 is
+"fine for W1."
+
+## What I verified, not assumed
+
+- `crates/sim-host/src/pacer.rs`'s doc comment on `wait_for_next` already carries a
+  `CORRECTION (issue #168)` block describing the real behaviour (bounded catch-up bursts, correct
+  on average, deliberately not reset-to-now) — ask 1, done.
+- `Pacer::jitter_percentiles()` (nearest-rank p50/p99/max over a bounded 2,000-sample window) is
+  implemented and covered by dedicated tests (`jitter_percentiles_are_empty_before_any_call`,
+  `on_time_ticks_report_zero_jitter`, `one_late_tick_among_on_time_ones_shows_up_in_max_but_not_p50`,
+  `the_jitter_window_forgets_samples_older_than_its_capacity`), and `host.rs`'s stats file /
+  `wire-probe`'s report line both carry the percentiles alongside the existing miss count — ask 3,
+  done. Landed in `120c3dc` (PR #267, 2026-08-08), which itself records "asks 2 and 4 are
+  explicitly deferred by the issue itself" — but the PR's own commit message and its
+  `roles/senior-controls/log/2026-08-08-issue-168-jitter-percentiles.md` entry both state
+  up front that the PR "references #168, does not close it," and no later PR added a `Closes
+  #168`. So the issue sat open with both in-scope asks already landed, same shape as #230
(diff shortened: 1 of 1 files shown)


MikePaNtZ:
The `rust` check failure here is unrelated to this PR's diff (one new Markdown file, zero Rust changes): CI's floating `dtolnay/rust-toolchain@stable` picked up a new clippy lint (`clippy::chunks_exact_to_as_chunks`, clippy 1.98.0) that fires on pre-existing code in `crates/plant-mujoco/src/bin/replay.rs`. It will fail identically on any branch built off current `master` right now, including this one.

Fixed separately in #292 rather than widening this PR's scope. Once #292 lands, `master` moves and this branch's required checks will pick up the fix on its next update — no push needed here.

---

Issue #235, opened 2026-08-06
Braking authority is battery-state-dependent — a full pack cannot absorb regen, and nothing in our design accounts for it


MikePaNtZ:
## What

Braking on this class of board regenerates into the battery. **A full pack cannot absorb regen, so braking authority falls as state of charge approaches 100%.** This is not a hypothesis — it is manufacturer-documented behaviour on production Onewheels.

Future Motion ships **"Full-Battery Pushback"**: at high SOC the board lifts the nose to create resistance and limits further braking, because it has nowhere to put the energy. Their own guidance is explicit: *never ride down a long or steep hill on a freshly charged board.* The documented workaround is to power-cycle and climb briefly to burn off charge before descending.

Threshold is approximately 90–100% SOC (the ~90% figure comes from their companion app's pre-descent notification and community synthesis, so treat the exact cutover as approximate rather than manufacturer-certified).

## Why it matters to us

**The worst-case braking condition is top of a hill on a fresh charge** — exactly when the rider most needs to stop. Our model has no battery, no SOC, and no regen limit at all. Every braking number we have been measuring assumes the pack will absorb whatever we send it.

This is a second, independent, state-dependent ceiling sitting *below* the motor-torque bound under the conditions that matter most. It is invisible to every test we currently run.

It also bears directly on the reference-governor design now being scoped. A governor computing an achievable deceleration profile from a fixed torque ceiling would be **wrong precisely in the scenario the manufacturer warns about**, and wrong in the unsafe direction — promising a deceleration the pack cannot accept.

## Related context — the failure this class of defect produces

The 2023 CPSC recall (~300,000 units, at least 4 deaths) is centred on the board ceasing to balance the rider when limits are exceeded. The litigation's engineering allegation (MDL 3087, N.D. Cal. — allegations, not adjudicated fact) is structurally relevant: pushback is alleged to draw on the same depleting motor/battery resource whose exhaustion causes the nosedive, so the warning degrades at the moment it is needed. The 2023 Haptic Buzz remedy is alleged to inherit the same flaw.

Design rule this implies, beyond the SOC question: **a limit warning must not share fate with the resource whose exhaustion it warns about.**

## Asks

1. Decide whether the sim needs a battery/SOC model at all for Stage 0, or whether the regen ceiling can be represented as a parameterised limit on regenerative torque without modelling the pack.
[...]

Issue #232, opened 2026-08-05
Tilted-ground and rotated-gravity hill formulations disagree by 3.5% and no solver setting fixes it


MikePaNtZ:
## What

`sim/scenarios/plant.py` offers two ways to model a hill: tilt the ground, or rotate gravity. They are **physically identical** for a point mass, and the repo's own test asserts they agree within **2%**.

At 10% grade they disagree by **3.48%**, and the test fails.

## It is not a convergence problem

Swept as part of an unrelated artefact investigation:

| knob | effect |
|---|---|
| solver iterations 100 → 200 → 500 | **zero** — traces bit-identical |
| tolerance → 1e-12 | **zero** — traces bit-identical |
| timestep ×0.5, ×0.25, ×0.1 | error moves **non-monotonically in a 2.86%–3.48% band, never below 2%** |

No tested setting achieves compliance. Newton is already fully converged, so this is not numerical slop that finer settings will resolve.

## Likely cause

A structural difference between the two formulations once the wheel has **finite contact geometry** rather than being a point mass. Rotating gravity leaves the contact patch orientation unchanged relative to the wheel; tilting the ground does not. For a point mass this distinction vanishes, which is why the equivalence reads as exact on paper.

Adding a constant Coulomb `frictionloss` term (`feat/mech/drag-model`) may interact with this, since constraint-based friction resolves against the contact frame — but the disagreement should be characterised on its own terms first.

## Why it matters

Scenarios choose between these formulations for convenience, and results are compared across them as though they were interchangeable. At 3.5% they are not. Any acceptance number that silently depends on which hill model was used carries that error, and nothing currently flags it.

## Asks

1. Characterise the disagreement versus grade — does it grow with angle? Does it vanish at 0%?
2. Determine whether one formulation is more correct for a finite-radius wheel, and say which scenarios should use which.
3. Either fix the equivalence or replace the 2% assertion with a documented, derived bound — but do not simply widen the threshold to whatever currently passes.

Owner: Sr. Mechanical & Systems (the sim-to-hardware fidelity contract). Surfaced during the ADR-0011 artefact battery; explicitly out of scope of that launch verdict and unrelated to it.

Issue #208, opened 2026-08-02
Encode the authored-world constraint as a check -- condition 2 of the ADR-0011 criterion move


MikePaNtZ:
## Summary

ADR-0011's second ratification moved the kerb-strike criterion to the hardware gate. That move is **only honest if the authored world is constrained to what the controller actually survives, and the constraint is encoded as a checkable rule rather than a hope.**

This is condition 2 of three. If it is dropped, the criterion move becomes the softening manoeuvre the ADR forbids, and the hold does not clear.

## What the controller actually survives

Measured, at full stick:

- rides out roughly a **1 mm** lip at a calm point of the hold, and **nothing** at its worst point;
- a **20 mm** lip — brick edge, root heave, ordinary pavement — imparts **201 °/s** of pitch rate against a KD channel affording about **76 °/s** at zero lean;
- inverts at **±0.5°** of static pitch error, survives **±0.25°**. A static incline is an effective static pitch disturbance, so **authored slope eats directly into that band.**

This is a geometry and actuator limit, not a tuning gap. ADR-0011 concluded the undersized elements are **ballast stroke and CoM height**, not the motor. No software change moves these numbers.

## What this means for the world

The shipped world must be **analytically smooth** — flat planes and clean meshes, no step discontinuities at the millimetre scale that a mesh seam or a collision-hull artifact can easily produce accidentally.

Authored inclines must stay **well inside** the measured static tolerance. The exact tolerance number is being measured — it is predicted under 0.5° but that is a prediction, not a measurement.

## The actual ask

**Encode it, do not document it.** A convention in prose is a polling surface; this needs to be a check that fails when the world violates it. If terrain is data, this is a checkable asset constraint.

Something like: a validator over the shipped terrain assets that fails on any step discontinuity above a stated height, and on any authored slope above the measured tolerance.

## Acceptance criteria

1. A terrain asset violating the step-height or slope limit **fails a check**, demonstrated by authoring a violating asset on purpose and watching it fail. Reading the validator is not evidence it works.
2. The limits cite the measured numbers as provenance rather than restating them.
3. The check runs where terrain assets actually change, not only where control code does.

## Dependency

The incline tolerance number comes from the controls-side sweep. **The step-height half can start now** — the ~1 mm figure is already measured.

## Boundary note

The board physics stay in `overboard`; nothing outside it computes board state. This is an asset-validation rule consuming numbers measured there, not a re-derivation of them.

## The other two conditions, for context

[...]

Issue #204, opened 2026-08-02
Braking is far too weak and coasting far too free -- regen and rolling resistance are both missing


MikePaNtZ:
## Summary

Braking is far too weak and coasting is far too free. Reported from hands-on driving. Neither regenerative braking nor rolling resistance exists in the model at all — this is not a tuning gap, it is two missing terms.

## What a rider expects

A real onewheel-class board brakes **hard** through regen, and it has real rolling resistance so it sheds speed when you stop commanding. The current board does neither: it holds speed when the stick returns to centre, and asking it to slow produces far less deceleration than the same input on real hardware would.

The gap shows up as the board feeling like it is on ice — which is the single most-reported thing after the flip.

## Where it lands

The fore-aft command path, plus a drag term. Same region of `crates/sim-host/src/host.rs` (`run()`, ~lines 773–873) as everything else queued right now.

## Sequencing — this is the constraint, not the code

Three separate pieces of work all land in that one region:

1. the command-envelope reserve and loss-of-authority warning (`feat/controls/cmd-envelope-reserve`, in flight);
2. reset (#202) and turn radius (#201);
3. this.

**Do not run them in parallel.** Sequence through the file. This one should go last of the three, because it is the only one that changes what the board *does* rather than what it is *allowed* to do, and it wants a settled baseline to be judged against.

## The trap worth naming up front

Braking loads the board in the same direction gravity does during a forward stop. ADR-0011 established that **the survivability boundary is the speed cap, not the lean** — saturation is survivable if and only if something is already unloading the board when it occurs — and that the worst untested case is a **full reverse-to-forward stick reversal at speed**, where commanded deceleration and gravity load the same side.

Stronger braking moves directly into that case. It is entirely plausible that adding realistic braking authority creates a *new* way to invert the board, in the exact regime the launch is being held over.

So this is not the cosmetic item on the list. It needs the ADR-0011 criterion (a) matrix re-run after it lands, not just a feel check.

## Acceptance criteria

1. Deceleration under full brake command is stated as a **measured** number in m/s², with the stopping distance from a stated speed. Not "feels better".
2. A rolling-resistance term exists with a derived value and a provenance comment, in the style the repo uses for derived constants. Deriving it from the board's mass and wheel geometry beats picking a number that feels right — an undocumented constant is how `damping="0.08"` became a blocker.
[...]

PR #295, opened 2026-09-15
Re-verify ADR-0011's (f1)/(f2) trim pin and condition-2 inputs against the corrected drag model

MikePaNtZ:
## What changed

No open issue was cleanly dispatchable this pass (detail in the log entry below), so this
addresses a decision-record correctness gap found while checking: **ADR-0011's third
ratification** (PR #237, 2026-08-06) declared every quantitative claim in the ADR predating it —
including the **(f1)/(f2) trim pin** and the incline-tolerance/kerb-strike numbers **condition 2**
depends on — superseded and uncitable *"until re-measured against the corrected model."* Nobody
had recorded that re-measurement in the ~30 PRs merged since, and `roles/senior-controls/CONTEXT.md`
(read by every Senior Controls session first) still frames the **second** ratification as current
and calls issue #208 (the condition-2 asset validator) **"not my turf"** — which `policy_check.py`
shows is simply wrong, and which had already caused one prior cron pass (2026-08-16, logged in
this same directory) to skip #208 for that stated reason.

**Did the re-measurement for real**, not by assumption: built the exact pinned toolchain
(Python 3.12 + `mujoco==3.10.0` + `numpy==2.5.1`, since this sandbox's default `python3` is 3.11)
and `sim-host --release`, then ran `tests/test_cmd_envelope_reserve.py` and
`tests/test_incline_tolerance.py` against current master. Every pin reproduces to the precision
already asserted:

- (f1) trim measured **−2.5009°** against the **−2.501° ± 0.10°** pin
- (f2) peak-demand slope **41.967 A/unit**
- criterion (c) warning lead **2.868 s**
- kerb-strike **20 mm / 201°/s** (still a strict xfail — moved to the hardware gate)
- incline sweep: no inversion anywhere swept; corridor-brake arrest boundary **6.5° arrested / 7.0° outrun** (exact match to prior figures)
- residual ratio (the atan(a/g) ≈ 1 rad/g identity (f) depends on): **1.028** — did not move

Per ADR-0011's own branch for this case ("if the slope moved, re-derive; if not, …"), **no
re-derivation is triggered** — the pins stand, now with a dated, recorded re-verification instead
of a silent assumption that CI passing meant the same thing. Full workspace build + full suite
for a clean baseline: **348 passed, 7 xfailed, 0 failed**.

## Why

ADR-0011 is a launch hold; its exit conditions exist specifically so nothing ships on an
unverified claim. The third ratification explicitly voided the numbers condition 2 needs, and
the role's own standing context — the document every future session reads first, before the
ADR itself — never caught up, to the point of actively steering a prior session away from the
one open issue that's actually workable. Left uncorrected, the next session either re-skips
#208 for the same wrong reason, or builds it citing numbers the ADR says cannot be cited without
this check having been done.

## What this satisfies

- Clears ADR-0011 third ratification's "re-measured against the corrected model" gate for
  (f1)/(f2) and for condition 2's input numbers.
- Corrects `roles/senior-controls/CONTEXT.md`'s stale ratification pointer and its wrong "#208
  not my turf" line (verified against `policy_check.py`, not asserted).

## Deliberately left out

- **#208 itself** (the terrain/asset validator script) — now unblocked, but writing the
  validator + a deliberately-violating fixture + CI wiring is its own scoped unit of work and
  would have pushed this past one thing. Flagged for the next Senior Controls pickup.
- **#204, #235, #232** — each blocked (sequencing / open design question / underspecified) for
  reasons detailed in the log entry, not re-litigated here.
- **The rest of `CONTEXT.md`** — its Stage-0B section is separately, significantly stale (describes
  PR #197 as not building against a master that has since shipped a working, hardware-verified
  Pi image); out of scope for a pass about the ADR-0011 balance-loop thread specifically.

Full detail, every number, and the reasoning for skipping each of the 23 open issues:
`roles/senior-controls/log/2026-09-14-adr-0011-third-ratification-reverify.md`.

No issue number to close — this is a decision-record correctness finding raised directly by
ADR-0011's third ratification, not a numbered issue.

https://claude.ai/code/session_018o8LUG5ihyAJMVuGzhX2hg

---
_Generated by [Claude Code](https://claude.ai/code/session_018o8LUG5ihyAJMVuGzhX2hg)_


MikePaNtZ:
`publish-sim-artifact` is failing on this PR (step "Verify the terrain clips were actually produced"): `scripts/render_scenario.py --scenario terrain --compare --source sim` asserts the estimator-attitude run does **not** reach the next crest and **does** nose-strike, so the published clip shows a real truth-vs-estimate divergence — but on this tree it now completes the ride instead.

**Not this PR's** — this branch's diff is docs/log-only (`roles/senior-controls/CONTEXT.md`, a new log file, and two test-docstring provenance notes; no `sim/`, `scripts/`, or `crates/` code changed). Confirmed rather than assumed: master's own current HEAD (`787c664`, exactly what this branch is based on) fails **identically**, same step, same assertion — [run 32611551986](https://github.com/MikePaNtZ/overboard/actions/runs/32611551986/job/97125755673). So this is a pre-existing, deterministic condition on `master`, not something this PR introduced or can fix without expanding well past its scope (this PR is specifically the ADR-0011 balance-loop re-verification pass).

No fix ported: one was attempted for this same comparison before (PR #264, `re-derive the CI terrain-comparison render's grade too`) and was explicitly reverted in the same PR, so there isn't a standing fix to carry over — re-deriving the terrain-comparison envelope is its own piece of controls work, not a rebase-and-port.

`publish-sim-artifact` is deliberately not a required check (CLAUDE.md: "it only runs on push to master, so requiring it would hang every PR forever") — the render/validate half does also run on PRs, but merge is gated on `rust`/`sim` only, both green here. Flagging this rather than re-running it: a re-run would reproduce the same deterministic assertion failure, not a flake. Left as a known, open, unrelated issue for a future controls pass to re-derive the terrain-comparison envelope (same shape as #250's fix to this file).

---

